PR #8, opened 2019-01-02
Broke out Ballot, Contest, and Printable into separate files

mirmasoodali:
This is an initial fix. Further granular breaking out will be necessary for increased modularity.


beausmith:
@mirmasoodali Thanks for this PR. I feel bad that we were not able to use it. If you pull `master` you'll see that the app has been re-written in Typescript. We'll be cleaning up the issues and hopefully you find some you're interested in working on. Please comment on the issue if you'd like to note that you're working on it or if you have any questions before you begin. Thanks again!

PR #13, opened 2019-01-17
Typescript app

beausmith:
Ok! First PR here. Take it for a ride. 

Looking forward to your comments and questions.


package.json
```diff
@@ -1,20 +1,39 @@
 {
-  "name": "bmd",
+  "name": "bmd-typescript",
```
benadida: do we need to call it `-typescript`?

public/index.html
```diff
@@ -19,13 +22,11 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>VotingWorks Ballot Marking Device</title>
+    <title>React App</title>
```
benadida: couple small things lost like the title of the page, but not a big deal, we can fix later.

package.json
beausmith: pushed fix.

public/index.html
beausmith: pushed fix.

PR #18, opened 2019-01-17
Consistently return to start page on ballot reset.

### LINKED ISSUE #17: Resetting ballot doesn't always go back to the Start Page

beausmith:
Fixes #17

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -29,6 +29,9 @@ const initialState = {
 }
 
 class App extends React.Component<Props, State> {
+  public initialWindowHistoryLength =
+    window.location.pathname === '/' ? window.history.length : 0
+
   constructor(props: Props) {
     super(props)
     this.state = initialState
@@ -36,6 +39,9 @@ class App extends React.Component<Props, State> {
 
   public resetBallot = () => {
     this.setState(initialState)
+    window.history.go(
+      -(window.history.length - this.initialWindowHistoryLength)
+    )
   }
 
   public updateVote = (contestId: string, vote: Vote) => {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -29,13 +29,19 @@ const initialState = {
...
+    window.location.pathname === '/' ? window.history.length : 0
+
   constructor(props: Props) {
     super(props)
     this.state = initialState
   }
 
   public resetBallot = () => {
     this.setState(initialState)
+    window.history.go(
```
benadida: what's the reason for making this complicated? I would have done

`window.location.replace('/');`

beausmith: Good question: because this will only replace the current location with the new location. This, if a back button (or action) is present, then a voter will be able to still go back to previous screens. 

Without this fix, if you open a new window, enter the app url, then complete the ballot, then click the "start over"… you are taken back to the newly opened window or the page that linked to the app. 

This change will take the voter back to the initial start page from when the app first loaded. 

I know that these scenarios may not be relevant once the app is on a device. Until then, the demo app is available on the web and should not have bugs in the web env.

benadida: Agreed, we shouldn't have bug in web just because it's in a web browser. And so I see your point.

That said, I would then strongly advocate for navigation from question to question to just not alter the URL. Back and Forward buttons simply don't navigate within the ballot. I think if we allow back and forward button, we'll have to start considering a number of complicating scenarios that simply won't matter in the real setting.

What do you think?

beausmith: Ok. Let's create an issue for replacing the url-based navigation with state-based navigation. 

This PR is meant to be a bug fix for #17.

benadida: Ok sounds good.

Issue #24, opened 2019-01-23
Use localStorage for persisting loaded election config data


beausmith:
- [x] Save data from `election.json` into localStorage
- [x] Load election from localStorage in app index.
- [x] Provide action to clear localStorage. This action can be disabled in production mode as one of many ways to not allow election config to change once loaded.

beausmith:
Done

PR #25, opened 2019-01-23
BMD can be configured with JSON election file

### LINKED ISSUE #15: BMD can be configured by adding a JSON election configuration file.

beausmith:
Resolves #15


src/App.test.tsx
```diff
@@ -2,10 +2,8 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import App from './App'
 
-import election from './data/election.json'
-
 it('renders without crashing', () => {
   const div = document.createElement('div')
-  ReactDOM.render(<App election={election} />, div)
+  ReactDOM.render(<App election={undefined} />, div)
```
benadida: is this the right approach? My sense is that making this nullable introduces the chance for problems. Feels like either `App` should have no parameter and have the config upload capability within it, or `App` should have a parameter that is non-nullable and the config upload capability should be outside of it.

What do you think?

beausmith:
Demo: https://deploy-preview-25--vw-bmd.netlify.com/

src/App.tsx
```diff
@@ -50,8 +47,15 @@ class App extends React.Component<Props, State> {
     }))
   }
 
+  public setElection = (election: OptionalElection) => {
+    this.setState({ election })
+  }
+
   public render() {
-    const { contests } = this.props.election
+    if (!this.state.election) {
```
benadida: this is the kind of conditional that I worry about once the election is nullable. We lose some of the typescript goodness.

src/App.test.tsx
beausmith: Yes, I agree. I was considering making a sister to `App` component which is the `Config` component which is loaded when there is no election value in localStorage. Read the description in this issue and add your comments:

- Use localStorage for persisting loaded election config data #24

src/App.tsx
beausmith: Yes, I agree. I want to make sure that the election object is not part of any component state. See previous comment for future solution.

PR #27, opened 2019-01-29
Basic end to end test: WIP

### LINKED ISSUE #16: An end-to-end test confirms voter can complete ballot.
Test should:
- [x] start on start page.
- [x] advance to each contest and select candidate
- [x] view summary
- [x] click print (possible?)
- [x] reset back to start page.

beausmith:
This is not ready for full review… commits need to be cleaned up.

Pull and run `yarn test` to get watcher where you can interactively run updated/all tests… or `yarn test:precompile` to run all tests and exist.

Closes #16


benadida:
Overall looks really cool. I'm wondering about the snapshots: what is the balance of how useful they are to detect problems vs. how much work it will be to change as we add features. No need to change anything yet, just something I'm wondering.

beausmith:
@benadida - So we're looking pretty good. That last little bit… I'm not really sure how to test it, but I think we can figure it out later. 

image

The commits are a bit of a mess… but will be easy to clean up. 

This branch has a conflict as I need to rebase on master and fix some conflicts.

Please poke around and LMK if you have any questions or comments.

beausmith:
If I deem that the context file is not necessary to test… and I add it to the ignored files config… then we're 100% green!

image

src/pages/SummaryPage.tsx
```diff
@@ -31,17 +30,18 @@ const TableCell = styled.td`
...
 const SummaryPage = (props: RouteComponentProps) => {
   const { contests, resetBallot, votes } = useContext(BallotContext)
-  const startOver = () => {
+  const getNewBallot = () => {
     if (
       Object.keys(votes).length === 0 ||
       window.confirm('Clear all votes and start over?')
     ) {
       resetBallot()
+      props.history.push('/')
```
benadida: what's the thinking behind separating the navigation out of `resetBallot()`? Is there a chance we would reset ballot without returning to the home screen and is that bad?

beausmith: 👍  Yeah, I noticed this too. I'd like it to be part of the `resetBallot()` function. Trouble is that the `history` prop is only available within the router. But… when we get to the task of replacing the router with a state-based navigation, then we can do this. So, it'll certainly get refactored then. 

I'm gonna refactor the "wip" commits out of this branch and then I'll be ready to merge it. Looking forward to getting into the EMS!

benadida: 👍

PR #36, opened 2019-02-13
progressive web app

benadida:


benadida:
@beausmith good call, I don't think I mind the logging for now, but will be good to remove later. Tracking in #41. 

Working on updating graphics now.

PR #40, opened 2019-02-14
Name as id (anticipation of write-in-candidates)

beausmith:


benadida:
Let's discuss this, I'm pretty sure we need an identifier for the non-write-in candidates, so that we have short-code ways of identifying all non-write-in candidates (LA County does this, for example, on their ballot.)

benadida:
A little more detail on my quick comment this morning: write-ins are usually tallied totally separately, in fact they're often not tallied unless the total number of write-ins might actually influence an outcome, so they *should* be coded separately.

The reason we still want an `id` for non-write-in choices is that we have to produce Cast Vote Records (CVRs) that need to be compatible with older systems, sometimes old Excel spreadsheets, and the last thing I want us to have to do is have a name, which might include some non-ascii characters, be used as an identifier in these clunky old systems. We need an ASCII-only identifier for every candidate.

beausmith:
Thanks @benadida. So can I assume that the EMS will create these ASCII-only ids from the Candidate Name and that the id should not be updated once created in the JSON?

benadida:
@beausmith EMS should allow the ID to be entered by election officials, some of them like very short ones, e.g. `HC` or `DT` for `Hillary Clinton` and `Donald Trump`. But auto-generation in the EMS UI is cool.

Absolutely right, no modification once the `election.json` comes out of EMS.

beausmith:
@benadida What VotingWorks app will create the Cast Vote Records (CVRs)?

benadida:
I should diagram this out!

The `election.json` will come from EMS. It will be consumed by the BMD and the BRD (Ballot Reading Device), i.e. the scanner. The scanner produces CVRs.

benadida:
Some of this is described in https://docs.google.com/document/d/1vAe1-I3QFvSF2dxeTP7bAA_qVDV1P4OY1rA7mcuwORg/edit

PR #49, opened 2019-02-20
Local storage

### LINKED ISSUE #48: Use localStorage for election data.
- [x] read data from localStorage if exists.
- [x] configure stores data upon configure.
- [x] clear data via keyboard combo.

beausmith:
Fixes #48 

So the functionality works, but I can't get the test to fire a keyboard event that Mousetrap will "hear". See commented out bits in the last commit.


beausmith:
I will rebase this PR off the `anywhere-ballot-styles` branch tomorrow.

src/pages/SettingsPage.tsx
```diff
@@ -23,6 +44,13 @@ const SettingsPage = (props: RouteComponentProps) => {
         <div />
         <div />
       </ButtonBar>
+      <Modal isOpen={showResetBallotAlert}>
+        <Text>Are you sure you want to remove all votes and start over?</Text>
```
benadida: change text from `remove all votes` to `clear all selections`

src/pages/__snapshots__/SettingsPage.test.tsx.snap
```diff
@@ -78,6 +90,18 @@ exports[`renders SettingsPage 1`] = `
       <p>
         Settings will be available here.
       </p>
+      <h2>
+        Clear Ballot Votes
```
benadida: change to `Clear Selections`

src/App.tsx
```diff
@@ -25,17 +40,37 @@ const initialState = {
...
+    Mousetrap.unbind(removeElectionShortcuts)
+  }
+
+  public getElection = () => {
+    const election = window.localStorage.getItem(electionKey)
+    return election ? JSON.parse(election) : undefined
+  }
+
+  public setElection = (election: Election) => {
+    this.setState({ election })
```
benadida: for increased defensive coding, I'd suggest having a `reset(election)` method that both sets the `localStorage` and React component state. In other words, minimize the touchpoints between your code and `localStorage` as much as possible.

src/App.tsx
```diff
@@ -45,8 +80,10 @@ class App extends React.Component<{}, State> {
     }))
   }
 
-  public setElection = (election: Election) => {
-    this.setState({ election })
+  public resetVotes = () => {
```
benadida: is `resetVotes` meant to be resetting a single ballot's selections? If so, strongly recommend naming this `resetSelections` or `resetBallot`, otherwise it sounds like you're tracking the set of votes cast and this is how you remove all the votes.

src/App.tsx
beausmith: I'm not sure what you mean. I think a diff here would more clearly explain your suggestion.

src/pages/SettingsPage.tsx
beausmith: fixed

src/pages/__snapshots__/SettingsPage.test.tsx.snap
beausmith: fixed

src/App.tsx
beausmith: Sure, we can update this. I don't think we all have a shared understanding of what all the various terms mean. If the device doesn't have a feature to track all votes cast, then a function `resetVotes` would not be doing anything other than clearing the votes/selections of the current voter. At present this function resets the `votes` key to an empty object, thus the function clearly describes what it is doing. Let's do a full pass across all the function names when we're together next to make them as clear as possible.

beausmith:
@benadida - this branch has been rebased, is passing, and is ready for review.

There is one part that I've been stumped on… I can't get the test code to fire and event which Mousetrap can listen to and trigger the `removeElection` method.

I have tried alternate libraries to Mousetrap, but nothing changes.  I will highlight the code to uncomment to see the tests which are not passing (and currently ignored) but that should pass.

src/App.test.tsx
```diff
@@ -3,16 +3,82 @@ import ReactDOM from 'react-dom'
...
+
+    // Tests are not passing.
+    // Unsure why keyboard event does not trigger Mousetrap.
+
+    // Basic "react-testing-library" keyboard event:
+    // fireEvent.keyDown(document.body, {
+    //   code: 'KeyK',
+    //   key: 'k',
+    //   metaKey: true,
+    // })
```
beausmith: Uncomment this `fireEvent` block

src/App.test.tsx
```diff
@@ -3,16 +3,82 @@ import ReactDOM from 'react-dom'
...
+    // document.dispatchEvent(
+    //   new KeyboardEvent('keydown', {
+    //     keyCode: 75,
+    //     metaKey: true,
+    //   } as ModifiedKeyboardEventInit)
+    // )
+
+    // Tests to confirm keyboard shortcut was pressed…
+    // await waitForElement(() => getByText('Configure Ballot Marking Device'))
+    // expect(window.localStorage.getItem(electionKey)).toBeFalsy()
```
beausmith: Uncomment these two lines, which should pass.

src/App.tsx
```diff
@@ -1,6 +1,25 @@
...
+//     charCode: event.charCode,
+//     code: event.code,
+//     key: event.key,
+//     keyCode: event.keyCode,
+//     metaKey: event.metaKey,
+//   })
+//   console.log('==============================')
+//   console.log('==============================')
+//   console.log('==============================')
+// })
```
beausmith: Uncomment this whole `document.addEventListener` in order to view the event in:
- the console from manual keyboard shortcut entry.
- the test output when running `yarn test` in your terminal.

src/App.tsx
beausmith: Doing a bit of refactoring… and I'm convinced. Changing `resetVotes` to `resetBallot`.

benadida:
@beau I suspect the issue is with web content sandboxing: it's not possible to truly simulate actual keypresses in web content Javascript. I think you have to test mousetrap like this:

https://craig.is/killing/mice#api.trigger

benadida:
bug: if I select a choice for question 1, press back, ctrl-K to back out to election definition file, choose sample election, click get started, my selection in the ballot is maintained. Should do a hard state reset upon loading an election file.

src/App.tsx
benadida: hard for me to do a diff right now, am sneaking in a few minutes of review here and there.

Let me try to rephrase what I mean.

Right now `removeElection` and `setElection` both touch `window.localStorage` and call `setState`. Feels a bit flimsy for both to have to coordinate those two pieces of state, best if that coordination lives in only one place. One way to address this: `removeElection` just calls `setElection` with a null election. There are other ways, I don't have a strong preference, only saying that the double state-setting -- `localStorage` and React state -- should happen in only one place, with other touchpoints using a higher-level abstraction.

beausmith: Thanks for this further explanation. Very helpful.

Though, I feel like it conflicts with a previous goal you had mentioned: that `setElection` should only accept an `Election` typed object. 

I have refactored the `removeElection` method to be named `reset`.

While both `setElection` and `reset` call `setState`, they both call a different `localStorage` methods, `setItem` and `removeItem`. Is this different enough to have two "touchpoints"?

beausmith:
Cool. I feel that I have addressed all your suggestions via code or comment. Happy to iterate further if you feel further changes are necessary.

Merging…

PR #52, opened 2019-02-22
Write in Candidates

### LINKED ISSUE #51: Support write-in candidates
- [ ] Clicking "Write-In Candidate" launches modal
- [ ] Modal has virtual keyboard with limited keys
- [ ] Stores to votes as `"custom": "ROGER RABBIT"`

beausmith:
Resolves #51 

Tests will fail because there is not 100% test coverage for this feature yet. I'm considering a temporary lowering of the code coverage percentages to be ready for demo. 

This feature is nearly complete. It needs a review and also [Simple Keyboard](https://franciscohodge.com/projects/simple-keyboard/documentation/) has many options which we should implement.

### DIFF
--- a/src/components/SeatContest.tsx
+++ b/src/components/SeatContest.tsx
@@ -114,7 +114,7 @@ class SeatContest extends React.Component<Props, State> {
     const { contest, vote } = this.props
     const { attemptedCandidateSelection } = this.state
     const selectedCandidate = contest.candidates.find(
-      candidate => candidate.id === vote
+      candidate => candidate.name === vote
     )
     return (
       <React.Fragment>
@@ -130,7 +130,7 @@ class SeatContest extends React.Component<Props, State> {
           </Legend>
           <Choices>
             {contest.candidates.map((candidate, index) => {
-              const isChecked = candidate.id === vote
+              const isChecked = candidate.name === vote
               const handleDisabledClick = () => {
                 if (vote && !isChecked) {
                   this.handleChangeVoteAlert(candidate.name)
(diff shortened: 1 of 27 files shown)


src/components/SeatContest.tsx
```diff
@@ -63,59 +66,156 @@ const Choice = styled('label')<{ isSelected: boolean }>`
...
+  writeInCandateModalIsOpen: false,
+  writeInCandidateName: '',
 }
 
 class SeatContest extends React.Component<Props, State> {
-  public state: State = initialState
+  private keyboard: React.RefObject<Keyboard>
+  constructor(props: Props) {
+    super(props)
+    let writeInCandidateName = ''
```
benadida: are we using empty string as the null write-in candidate name? That freaks me out a little bit. Should we not simply make this nullable?

src/components/SeatContest.tsx
```diff
@@ -130,54 +230,127 @@ class SeatContest extends React.Component<Props, State> {
...
               const handleDisabledClick = () => {
                 if (vote && !isChecked) {
                   this.handleChangeVoteAlert(candidate.name)
                 }
               }
               return (
                 <Choice
-                  key={candidate.id}
-                  htmlFor={candidate.id}
+                  key={candidate.name}
```
benadida: is the change from `id` to `name` because of write-ins? Given how these things are tallied, I suggest instead that we keep `id` as the unique identifier in every case, and use `[WI]` or some other write-in specific ID for the write-in option.

beausmith: There is no tally. This hasn't been a defined feature. Let's convert to IDs or some better option when we have this feature defined.

src/components/SeatContest.tsx
beausmith: Hmmm… Typescript is config'd to never use `null` only `undefined`. It's simpler Typescript code if types to not change.

Can you provide repro steps to show how things could go wrong?

Perhaps we `trim()` the string upon submission to catch any string containing spaces? 

Otherwise, I don't see how an empty string is any worse than `undefined‘.

beausmith: Cool. I've added some checks. Pushing new code shortly.

src/components/SeatContest.tsx
benadida: @beausmith I understand there's not tally here, but I still think this is a bad design approach, because I think identifiers should be different than visible names, especially as names might have multiple expression in different character sets. Let's switch this back to `id` please.

beausmith: I'd like us to discuss this more. I know the feeling you're having, I've gone back and forth, and I took this approach as it dramatically simplifies the logic for us and I can't think of a demonstrable reason not to do it. 

Let's merge this so as not to delay before demo and discuss early next week over video or in person and make a plan.

beausmith: Note: the IDs are still in the election.json.

beausmith:
Tests in we're still at 💯 🎉

PR #54, opened 2019-02-22
Activation Page (aka QR Code page)

### LINKED ISSUE #53: Add "Scan QR Activation Code" screen
- [x] the "Scan QR Code" screen is the first screen.
- [x] advance to the next screen by:
  - [x] tap the QR code on the screen
  - [x] scanning a QR code "MyVoiceIsMyPassword"

beausmith:
Not ready for merge, due to some tests not working quite right, but would be great to have a review.

Resolves #53


src/pages/SelectBallotPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  return (
+    <Main>
+      <MainChild center>
+        <fieldset>
+          <Legend>
+            <label htmlFor="ActivationCode">
+              <Prose textCenter>
+                <h1>Load Your Ballot</h1>
```
benadida: change to `Scan Your Activation Code`. I don't think voters are going to think of their QR code as their ballot and it's going to be confusing to them, even if it includes the ballot style.

src/pages/SelectBallotPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+let resetActivationCode: number
+
+const StartPage = (props: RouteComponentProps) => {
+  const [activationCode, setActivationCode] = useState('')
+  const goToStartPage = () => {
+    props.history.push('/start')
+  }
+  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setActivationCode(event.target.value)
+    if (event.target.value === 'MyVoiceIsMyPassword') {
```
benadida: the scanner sends a newline by default, let's make sure we either ignore the newline or make it part of the scan value.

beausmith: Bette yet, I'll make it a proper`<form>` such that the new line causes form to submit.

src/pages/SelectBallotPage.tsx
beausmith: Totes. All language needs to be reviewed.

beausmith: Fixed.

src/pages/SelectBallotPage.tsx
beausmith: fixed.

PR #59, opened 2019-02-25
make sure barcode appears on printed ballot

benadida:
Picked `min-height` 10in to have barcode at the bottom but consider 0.5" margins top/bottom.


benadida:
going to be more conservative on height, and also on the display version, which scrolls off the screen right now.

PR #76, opened 2019-02-27
Improve text-to-speech

benadida:
These changes include a nasty `setTimeout` hack, I'd be happy if we could do without it, but it's the simplest hack I could think of to get Chromevox to start talking right away on every new page.

There are a few tidbits that will make this much nicer, like adding periods that will make chromevox pause. There's also an issue that remains where on the second test, the tab order is messed up. But it's starting to be reasonable!

[update] added said punctuation.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -33,7 +33,8 @@ legend {
 /* Foundational */
 html,
 body,
-#root {
+#root,
+.clickContainer {
   height: 100%;
 }
 
--- a/src/App.css
+++ b/src/App.css
@@ -36,6 +36,7 @@ body,
 #root,
 .clickContainer {
   height: 100%;
+  overflow: hidden;
 }
 
 /* Typography */
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,6 +52,10 @@ const initialState = {
   votes: {},
 }
 
+// a React reference to the div that we want to click on for voiceover on every screen
+// this will be used in the Root component, and referenced in the App when location changes.
+const clickContainerRef = React.createRef<HTMLDivElement>()
+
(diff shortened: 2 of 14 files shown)


src/App.tsx
```diff
@@ -66,6 +70,18 @@ class App extends React.Component<RouteComponentProps, State> {
...
     }
     Mousetrap.bind(removeElectionShortcuts, this.reset)
+
+    // setting a delay of 50ms does the following:
+    // - the click means that chromevox will now read the whole page
+    // - thanks to giving it 50ms of delay, any component that is autoFocused will get autofocused first
+    //   which means that when the user presses tab, it goes to the item that was previously autoFocused.
+    // - setting this to 0 is too short to get the focus to happen.
+    // - setting this to 500 (for example) causes chromevox to start speaking the focused item, which is not what we want.
+    this.props.history.listen((location, action) => {
```
beausmith: `location` and `action` are unused variables. Typescript is complaining.

src/App.css
```diff
@@ -37,6 +37,10 @@ body,
   height: 100%;
 }
 
+#clickContainer {
```
beausmith: Let's add a class to the ruleset above: `.clickContainer`

src/App.tsx
```diff
@@ -130,9 +146,11 @@ class App extends React.Component<RouteComponentProps, State> {
 }
 
 const Root = () => (
-  <BrowserRouter>
-    <Route path="/" component={App} />
-  </BrowserRouter>
+  <div id="clickContainer" ref={clickContainerRef}>
```
beausmith: Change this `id` to `className="clickContainer"`

PR #77, opened 2019-02-27
added aria label for state seal

benadida:
otherwise it's gibberish when read by chromevox


benadida:
@beausmith forgive me for merging a tiny thing without your review so that this thing doesn't speak gibberish with a real state seal.

PR #87, opened 2019-02-28
allow all config files that end in .json

benadida:


src/components/UploadConfig.test.tsx
```diff
@@ -15,9 +15,13 @@ it(`allows file to be uploaded via file input`, async () => {
   fireEvent.change(fileInput, {
     target: {
       files: [
-        new File([JSON.stringify(goodElectionFile)], 'election.json', {
-          type: 'application/json',
-        }),
+        new File(
+          [JSON.stringify(goodElectionFile)],
+          'beausfavoriteelection.json',
```
beausmith: 😆

src/components/UploadConfig.tsx
```diff
@@ -60,7 +60,7 @@ class UploadConfig extends React.Component<Props, State> {
   public onDrop = (acceptedFiles: File[]) => {
     if (acceptedFiles.length === 1) {
       acceptedFiles.forEach((file: File) => {
-        if (file.name === 'election.json') {
+        if (file.name.endsWith('.json')) {
```
beausmith: Didn't know about this string method. 💡

PR #90, opened 2019-02-28
Activation (via QR code) can be disabled.

### LINKED ISSUE #82: Activation Screen via config.json

beausmith:
Resolves #82


src/components/Ballot.test.tsx
```diff
@@ -1,15 +1,21 @@
...
 
+import electionDefaults from '../data/electionDefaults.json'
+import electionSample from '../data/electionSample.json'
+
 import Ballot from './Ballot'
 
 it(`can navigate all ballot pages`, () => {
-  const { container, getByText, getByTestId } = render(<Ballot />)
+  const { container, getByText, getByTestId } = render(<Ballot />, {
+    election: lodashMerge(electionDefaults, electionSample),
```
benadida: It seems not quite right that the test file would have to know to merge the defaults with sample election. Feels like that should be within an abstraction of the core code, otherwise the tests strongly depend on this particular behavior.

src/App.test.tsx
```diff
@@ -1,9 +1,13 @@
+import lodashMerge from 'lodash.merge'
 import React from 'react'
 import ReactDOM from 'react-dom'
 import { fireEvent, render, wait, waitForElement } from 'react-testing-library'
 
-import electionFile from './data/election.json'
-const electionAsString = JSON.stringify(electionFile)
+import electionDefaults from './data/electionDefaults.json'
+import electionSample from './data/electionSample.json'
+const electionAsString = JSON.stringify(
```
benadida: Same comment as other: this merge doesn't feel optimal in the tests

src/components/Ballot.test.tsx
beausmith: Agreed. I had done used this, then removed it for a reason that has no merit. 🏆

beausmith: On second thought… I remembered why which has merit… but there is a better way! Amending previous commit.

src/components/Ballot.test.tsx
```diff
@@ -43,8 +46,22 @@ it(`can navigate all ballot pages`, () => {
...
+  const { container, getByText } = render(<Ballot />, {
     route: '/contests',
   })
+  getByText('President')
+  expect(container.firstChild).toMatchSnapshot()
+})
+
+it('skips activation if when disabled', () => {
+  const { container, getByText } = render(<Ballot />, {
+    election: lodashMerge(electionSample, {
```
benadida: being pedantic, I'd make defaults an optional argument to `mergeWithDefaults` and use it here. But this is optional, not required.

PR #94, opened 2019-03-01
Multiple seat support

beausmith:
This isn't finished… because tests aren't updated… but it works.


beausmith:
Here is a animated gif of the functionality…

multi-seat

benadida:
very impressed with how well this works with write-ins too.

src/components/SeatContest.tsx
```diff
@@ -236,17 +248,17 @@ class SeatContest extends React.Component<Props, State> {
...
-                {!!vote ? `1` : `0`}.
+                <strong>Vote for {contest.seats}.</strong> You have selected{' '}
+                {vote.length}.
               </p>
             </Prose>
           </Legend>
           <Choices>
             {contest.candidates.map((candidate, index) => {
-              const isChecked = !!vote && candidate.name === vote.name
+              const isChecked = !!vote.find(c => c.id === candidate.id)
```
benadida: here's that `find` again :)

src/App.tsx
```diff
@@ -113,9 +113,9 @@ class App extends React.Component<RouteComponentProps, State> {
     this.props.history.push('/')
   }
 
-  public updateVote = (contestId: string, candidate: OptionalCandidate) => {
+  public updateVote = (contestId: string, vote: OptionalVote) => {
```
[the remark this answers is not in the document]
beausmith: If confusing, let's update the terminology. Please propose any terms you find less confusing.

Here's my current thinking…

It hasn't been defined how a voter's "ballot" will be distilled from an election, but I've been thinking that `this.state.ballot` would be used for this. 

Possible `App` state:

```
{
  election: Election // <-- election config json
  ballot: Ballot // <-- TBD, created when an activation code is entered.
  votes: Votes {
    contestId: Candidate[] || Approval // <-- Approval will be a boolean type
  }
}
```

If the voter's "ballot" is dynamically derived from the election (instead of creating a ballot when an activation code is entered) then we can change `votes` to `ballot`.

src/components/SeatContest.tsx
```diff
@@ -125,101 +128,110 @@ class SeatContest extends React.Component<Props, State> {
...
 
-  public selectCandidate = (candidate: OptionalCandidate) => {
-    this.props.updateVote(this.props.contest.id, candidate)
+  public addCandidateToVote = (id: string) => {
+    const { contest, vote = [] } = this.props
+    const { candidates } = contest
+    const candidate = candidates.find(c => c.id === id)
+    if (candidate) {
+      this.props.updateVote(contest.id, [...vote, candidate])
+    }
```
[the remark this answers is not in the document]
beausmith: Hehe! Code coverage report (which I haven't completed yet) will complain that the else path is not followed. Ok I'll look into this.

src/components/SeatContest.tsx
```diff
@@ -125,101 +128,110 @@ class SeatContest extends React.Component<Props, State> {
...
-    const { value } = event.target as HTMLInputElement
-    const targetIsSelected = vote && value === vote.id
-    this.selectCandidate(
-      targetIsSelected
-        ? undefined
-        : contest.candidates.find(candidate => candidate.id === value)
-    )
+    const { vote = [] } = this.props
+    const id = (event.target as HTMLInputElement).value
+    const candidate = vote.find(c => c.id === id)
```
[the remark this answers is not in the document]
beausmith: Done.

src/components/SeatContest.tsx
```diff
@@ -125,101 +128,110 @@ class SeatContest extends React.Component<Props, State> {
...
+    if (!!candidate) {
+      if (candidate.writeIn) {
+        this.setState({ candidatePendingRemoval: candidate })
```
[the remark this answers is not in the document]
beausmith: If the voter took the time to add a write-in, and then accidentally clicks to uncheck and the write-in is gone without any warning confirmation… I think they would be disappointed. Thinking about older people fat-fingering the touchscreen. 😄

src/components/SeatContest.tsx
```diff
@@ -125,101 +128,110 @@ class SeatContest extends React.Component<Props, State> {
...
-  }
-
   public onKeyboardInputChange = (writeInCandidateName: string) => {
     this.setState({ writeInCandidateName })
   }
 
   public render() {
-    const { contest, vote } = this.props
+    const { contest, vote = [] } = this.props
+    const seatsAreFull = vote.length >= contest.seats
```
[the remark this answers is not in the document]
beausmith: Abstracted.

src/components/SeatContest.tsx
beausmith: Abstracted.

src/components/SeatContest.tsx
```diff
@@ -125,101 +128,110 @@ class SeatContest extends React.Component<Props, State> {
...
+    if (candidate) {
+      this.props.updateVote(contest.id, [...vote, candidate])
+    }
+  }
+
+  public removeCandidateFromVote = (id: string) => {
+    const { contest, vote = [] } = this.props
+    if (id) {
+      this.props.updateVote(contest.id, vote.filter(c => c.id !== id))
+    }
```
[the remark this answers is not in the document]
beausmith: Done and abstracted.

beausmith:
@benadida I believe I have addressed all of your comments. I'm clicking the link to request a re-review.

beausmith:
Looking into failed tests…

beausmith:
Fixed failing tests.

beausmith:
Now the tests are really fixed! 🤞

public/data/election.json
```diff
@@ -6,25 +6,37 @@
   "seal": "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 500 500\">\n  <circle cx=\"250\" cy=\"250\" r=\"240\" fill=\"#fff\" stroke=\"#000\" stroke-width=\"10\"/>\n  <text x=\"50%\" y=\"50%\" text-anchor=\"middle\" alignment-baseline=\"middle\" style=\"font-family: sans-serif; font-size: 180%\">Official Seal Here</text>\n</svg>\n",
   "contests": [
     {
-      "id": "president",
+      "allowWriteIn": true,
```
benadida: as per discussion, `allowWriteIns`.

src/components/SeatContest.tsx
```diff
@@ -125,101 +132,110 @@ class SeatContest extends React.Component<Props, State> {
...
     )
-    this.setState({ writeInCandidateName })
-    this.selectCandidate({
-      id: 'writeInCandidate',
-      name: writeInCandidateName,
-    })
+    this.props.updateVote(contest.id, [
+      ...vote,
+      {
+        id: camelCase(normalizedCandidateName),
```
benadida: I think the `id` here should probably be `write-in-1`, `write-in-2`, etc. I can imagine collisions between write-in names and existing candidates, and in general I think it may be a potential problem to use IDs that are decided by user input.

src/components/SeatContest.tsx
```diff
@@ -275,39 +292,58 @@ class SeatContest extends React.Component<Props, State> {
...
-          isOpen={!!attemptedVoteCandidateName}
+          isOpen={!!attemptedOverVoteCandidate}
           content={
             <Prose>
-              <Text>
-                To vote for {attemptedVoteCandidateName}, first uncheck the vote
-                for {!!vote && vote.name}.
-              </Text>
+              {contest.seats === 1 ? (
+                <Text>
```
benadida: nit: don't repeat what you don't need to repeat, e.g. `<Text>`.

beausmith: Simplified.

src/components/SeatContest.tsx
beausmith: Good call. Fixed by prefixing camelCased candidate name with `write-in__`.

beausmith:
@benadida Ready for re-re-re-review.

src/App.test.tsx
```diff
@@ -133,107 +129,147 @@ it('end to end: election can be uploaded, voter can vote and print', async () =>
...
-  fireEvent.click(getByText(contest0Candidate0).closest('label')!)
-  fireEvent.click(getByText('add a write-in candidate').closest('label')!)
-  fireEvent.click(getByText('Close'))
-  fireEvent.click(getByText('add a write-in candidate').closest('label')!)
+  expect(getByText(multiSeatCandidate0.name)).toBeTruthy()
+  expect(getByText(multiSeatCandidate1.name)).toBeTruthy()
+  expect(getByText(multiSeatCandidate2.name)).toBeTruthy()
+  expect(getByText(multiSeatCandidate3.name)).toBeTruthy()
+
+  // Test Write-In Candidate flow
```
benadida: awesome, full commitment to all test paths :)

src/components/Ballot.test.tsx
```diff
@@ -49,7 +49,7 @@ it('redirects contests index to first contest', () => {
   const { container, getByText } = render(<Ballot />, {
     route: '/contests',
   })
-  getByText('President and Vice-President of the United States')
+  getByText(electionSample.contests[0].title)
```
benadida: 👍

src/components/CandidateContest.tsx
```diff
@@ -125,101 +132,110 @@ class CandidateContest extends React.Component<Props, State> {
...
-    })
+    this.props.updateVote(contest.id, [
+      ...vote,
+      {
+        id: `write-in__${camelCase(normalizedCandidateName)}`,
+        isWriteIn: true,
+        name: normalizedCandidateName,
+      },
+    ])
+    this.setState({ writeInCandidateName: '' })
```
benadida: I'm still struggling with this empty string as null situation, but I'll leave it be until I have a better suggestion.

src/data/electionSample.json
```diff
@@ -554,37 +554,37 @@
...
...mber of any board, commission or other state or quasi-public entity exercising executive power under the laws of this state shall be a member of the senate or the house of representatives during his or her continuance in such office.\n\nb. Article V of the Constitution shall be amended to read as follows: The powers of the government shall be distributed into three (3) separate and distinct departments: the legislative, the executive and the judicial.\n\nc. Article VI, Section 10 of the Constitution shall be deleted in its entirety.\n\nd. Article IX, Section 5 of the Constitution shall be amended to read as follows:\n\nSection 5. Powers of appointment.- The governor shall, by and with the advice and consent of the senate, appoint all officers of the state whose appointment is not herein otherwise provided for and all members of any board, commission or other state or quasi-public entity which exercises executive power under the laws of this state; but the general assembly may by law vest the appointment of such inferior officers, as they deem proper, in the governor, or within their respective departments in the other general officers, the judiciary or in the heads of departments."
```
benadida: I think we decided not  to test the super long proposition right now. Are you sure you want to force this now? Feels like an edge case that could be tricky.

beausmith: Oh, I'm not aware of that decision. Please tell me more. 

My reasoning: It was in the sample content, so I assumed others were expecting it. Also long content is unavoidable unless we enforce it (when creating or loading json config). Since one of the next tasks is to add scroll buttons, I need long content anyhow. 

Happy to adjust.

PR #102, opened 2019-03-05
Typescript should return no errors when commiting.

beausmith:
Please approve this PR or https://github.com/votingworks/bmd/pull/101.


benadida:
rejecting in favor of giving a bit more freedom on branches, but #101 prevents merging bad code into master.

PR #105, opened 2019-03-06
Reset and reload app if error occurs.

beausmith:
@benadida - please do a post-merge review of this.


benadida:
late but still approve. Looks good, no comments.

PR #132, opened 2019-03-18
Cut the reset activation code time in half.

beausmith:
@benadida - Merge or close this PR as you see fit.


benadida:
On second thought I don't think this is necessary and I worry it will create some potential bugs. Let's kill it.

PR #136, opened 2019-03-19
Accessibility update

### LINKED ISSUE #92: Clean up accessibility
`setTimeout` is not your friend.

beausmith:
See commit messages for more details on changes.

Resolves #92


src/pages/CastBallotPage.tsx
```diff
@@ -49,9 +49,8 @@ const StartPage = () => {
       <MainChild center>
         <Directions>
           <Prose>
-            <h1>
+            <h1 aria-label={`Verify and Cast Your Ballot.`}>
```
benadida: I'm not excited about duplicating displayed and read content. That has all the risks of one changing without the other. Are we sure we want to adopt this pattern instead of the other?

beausmith: Yessir! Good question. I anticipated you would ask this. I was also on the fence about this for the same concerns you share. Here's my reasoning for where which side of fence I fell:

## Thoughts on `aria-label`:
- 😞 Because there are two strings to maintain, it's possible they may diverge.
- 😄 This is the purpose of `aria-label` and how I see it used in my research.
- 😞  When using the ChromeVox shortcuts, it's possible to get to a place where as the user tabs through, the voice over sees each hidden comma/period as a separate item when in reality it should just be a sentence w/o extra markup.
- 😄 When we localize, the string used will come from one source and be used twice, thus alleviating the duplication seen here.
- 😄 The `aria-label` strings are faster to read/grok/edit for humans.
- 😄 Less markup using `aria-label` than hidden elements. 
- 😄 Tests can be updated to look for specific strings in `aria-label`. Snapshots will also capture changes, but devs have to be aware of this.
- 😞 Strings with copious html markup are difficult to localize and difficult for translators to maintain.

## Thoughts on Hidden Elements
- 😄 Less string duplication, pre-localization
- 😞 Extra markup feels hacky compared to `aria-label`.

Considering all pros/cons, long-term my bet is that we will find that choosing `aria-label` will provide us with the best developer, translator, and voter accessibility experience.

benadida:
Lgtm!

Issue #139, opened 2019-03-20
Remove Yes/No Questions until YesNoContest component added.


beausmith:
Add removed json to #109

PR #140, opened 2019-03-21
remove yes/no questions for now

benadida:
we'll bring them back when we have a component to display them properly.


benadida:
@beausmith I'm getting a weird test failure locally of some duplicate ARIA ID, let me know if you see it too. But CI is passing, and I don't see any duplicate IDs, so I think this is an erroneous test report and I think this PR is fine.

beausmith:
Resolves #139

PR #145, opened 2019-03-22
Settings: font-size

beausmith:


beausmith:
@benadida - This PR is ready for your review. 

I created a second PR (https://github.com/votingworks/bmd/pull/152) to use the actual VVSG 2.0 font-sizes. This second PR serves as a reference implementation for others working on updating the design to work with the larger font sizes required by the VVSG 2.0.

src/App.tsx
```diff
@@ -110,9 +120,15 @@ class App extends React.Component<RouteComponentProps, State> {
   }
 
   public resetBallot = (path: string = '/') => {
-    this.setState({ votes: initialState.votes }, () => {
-      this.props.history.push(path)
-    })
+    this.setState(
+      {
+        ...initialState,
+        election: this.getElection(),
```
benadida: what's the reasoning for resetting the `election` state variable on `resetBallot`?

beausmith: TL;DR, this is simpler than keeping track of all the other keys.

I'm thinking that it's simpler to to reset all keys and then update/reset the election key, than to specify specific keys to reset here and to also remember to update these if/when additional keys are added to `initialState` in the future. 

What are your thoughts? Would you suggest another approach?

PR #147, opened 2019-03-26
upgrade packages and lock file to address recent security issues

benadida:


beausmith:
I'm curious about these recent security issues. I'm gonna google it… would be great to have a link in future PRs and/or commit msgs.

benadida:
Yeah I noticed them when playing with npm instead of yarn for the Cypress issue. Npm warns automatically. I believe yarn can be configured to do so, too. I have a task to follow up on it.

beausmith:
Loads of good stuff in this upgrade. TypeScript stuff too!

beausmith:
In the release docs they recommend using this command to upgrade:

    yarn add --exact react-scripts@2.1.8

Using `--exact` is new to me. Is this what you used @benadida?

case:
You might have a look at Dependabot (or something like it): https://dependabot.com/

It helps keep deps up to date. We use it and are huge fans.

benadida:
@beausmith no I used `npm audit fix` which updated the `package.json` file accordingly :) I'm happy to look into the right longer term solution.

beausmith:
I tried the recommended command `yarn add --exact react-scripts@2.1.8` and it produces one difference which may be worth making:

```diff
-    "react-scripts": "^2.1.8",
+    "react-scripts": "2.1.8",
```
By using the "exact" version, we are less likely to be surprised by any unexpected changes if yarn installs a newer version.

You decide, which PR you'd prefer: this one or https://github.com/votingworks/bmd/pull/151

benadida:
closing in favor of #151

PR #235, opened 2019-04-16
Adding ballot style support

benadida:
Modeling on real state data, we want to:
* group contests by districts: each contest points to a single district
* have a set of ballot styles, each referencing a set of districts.

Districts effectively serve as question groups, and we could name them that, but since they have a meaning in election administration, I'm choosing to explicitly call them districts right now. Think of congressional districts, state house districts, state senate districts, etc.

For completeness, we have a list of precincts that each ballot style might be found in. This is important for ballot activation where a precinct is chosen and then a ballot style.

TODO: add one more set of tests to show a ballot style that provides only a subset of questions.


benadida:
One thing I need to fix: right now my new tweak to how we handle errors causes a `setState()` inside a `render`, which causes a warning. I'm going to fix that with a separate `/reset` path that will do all the resetting outside of `render()`.

That said, I think this is otherwise ready for review @beausmith whenever you can.

src/pages/PrintPage.tsx
```diff
@@ -163,6 +163,14 @@ class SummaryPage extends React.Component<RouteComponentProps, State> {
...
     } = this.context.election
     const { showHelpPage, showSettingsPage } = bmdConfig
+
+    const contests: Array<CandidateContest | YesNoContest> = this.context
+      .contests
+
+    if (!contests.length) {
+      return <Redirect to="/reset" />
+    }
+
```
beausmith: If the only way to this page is by going through some contests… wouldn't `contests` always have a length?

src/pages/PrintPage.tsx
```diff
@@ -190,8 +198,8 @@ class SummaryPage extends React.Component<RouteComponentProps, State> {
               <Content>
                 <BallotSelections>
                   <BallotContext.Consumer>
-                    {({ election, votes }) =>
-                      election!.contests.map(contest => (
+                    {({ votes }) =>
+                      contests.map(contest => (
```
beausmith: Would probably be easier to follow code if context was accessed in one place, rather than two.

src/pages/ContestPage.tsx
```diff
@@ -20,38 +17,37 @@ const Progress = styled(Text)`
...
+  const { contestNumber } = props.match.params
+  const {
+    election,
+    updateVote,
+    votes,
+    contests: contestsOrNull,
+    resetBallot,
+  } = useContext(BallotContext)
+
+  const contests = contestsOrNull!
```
beausmith: This TypeScript project is configured to not allow the use of `null`, only `undefined`.

src/AppEndToEndAlternateBallotStyle.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+  c =>
+    c.title === 'Measure 102: Vehicle Abatement Program' && c.type === 'yesno'
+) as YesNoContest
+
+beforeEach(() => {
+  window.localStorage.clear()
+  window.location.href = '/'
+})
+
+it(`basic end-to-end flow`, async () => {
```
beausmith: Update this title to be more descriptive of what this test does.

src/AppEndToEndAlternateBallotStyle.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  // TODO: replace next line with "Enter" keyDown on activation code input
+  fireEvent.click(getByText('Submit'))
+
+  // Get Started Page
+  fireEvent.click(getByText('Get Started'))
+  expect(container.firstChild).toMatchSnapshot()
+
+  // make sure there are as many questions as we expect
+  getByText('1 of 11')
```
beausmith: Can we end the test right here? 

Is there anything after this which is specific to this test?

beausmith:
Pushed one more commit to trim/focus the test for Alternate Ballot Style.

src/App.test.tsx
```diff
@@ -49,14 +53,47 @@ describe('loads election', () => {
...
+
+    window.localStorage.setItem(electionKey, electionSampleAsString)
+    window.localStorage.setItem(
+      activationStorageKey,
+      JSON.stringify({
+        ballotStyleId,
+        precinctId,
+      })
+    )
+    window.localStorage.setItem(
```
benadida: It's very risky to store the voter's choice in `localStorage`. Too many ways that gets written to disk, not cleaned up. Same goes for the activation code, which might eventually include a signature. I feel pretty strongly that these should be only in memory, and should go away fairly aggressively if there's any reload / error / etc.

src/App.tsx
```diff
@@ -110,8 +78,21 @@ class App extends React.Component<RouteComponentProps, State> {
         election: mergeWithDefaults(electionSample as Election),
       })
     } else {
+      const election = this.getElection()
+      const { ballotStyleId, precinctId } = this.getBalotActivation()
```
benadida: typo, single `l` instead of double `l`.

src/components/Ballot.tsx
```diff
@@ -10,32 +10,35 @@ import HelpPage from '../pages/HelpPage'
...
+    showHelpPage,
+    showSettingsPage,
+  } = election!.bmdConfig!
 
   return (
     <Switch>
-      {requireActivation ? (
-        <Route exact path="/" component={ActivationPage} />
-      ) : (
+      {!requireActivation || (ballotStyleId && precinctId) ? (
```
benadida: I'm having trouble following this logic, and it looks error prone. For safety, error checks should, IMO, be closer to where the functionality is happening.

src/components/Ballot.tsx
```diff
@@ -10,32 +10,35 @@ import HelpPage from '../pages/HelpPage'
...
       )}
-      <Route exact path="/reset" component={ResetPage} />
       <Route exact path="/activate" component={ActivationPage} />
       <Route path="/cast" component={CastBallotPage} />
       <Route path="/start" exact component={StartPage} />
-      {contests.length && (
-        <Redirect exact from="/contests" to={`/contests/0`} />
-      )}
-
+      <Redirect
```
benadida: I don't like the style of changing the redirect based on a conditional. Conditional before the action is clearer.

src/components/YesNoContest.test.tsx
```diff
@@ -7,7 +7,7 @@ import YesNoContest from './YesNoContest'
 
 const contest: YesNoContestInterface = {
   description: 'description',
-  district_id: 'district-id',
+  districtId: 'district-id',
```
benadida: good call, my obsession with `_` is showing :)

src/config/types.ts
```diff
@@ -74,7 +75,7 @@ export interface Election {
   readonly ballotStyles: BallotStyle[]
   readonly precincts: Precinct[]
   readonly districts: District[]
-  readonly contests: Array<CandidateContest | YesNoContest>
+  readonly contests: Contests
```
benadida: good call, should have done that.

src/pages/ActivationPage.tsx
```diff
@@ -16,37 +18,45 @@ const CodeBox = styled.div`
...
   }
 
-  /* istanbul ignore next */
+  /* istanbul ignore next - shortcut will not exist in official release */
   const takeShortcut = () => {
-    setBallot(validBallotCodes[0])
+    const ballotStyle = election.ballotStyles[0]
+    setBallot({
+      ballotStyle,
+      precinct: election.precincts[0],
```
benadida: precinct should be amongst those that are valid for that ballot style. So:

`precinct: ballotStyle.precincts[0]`

src/pages/ActivationPage.tsx
```diff
@@ -16,37 +18,45 @@ const CodeBox = styled.div`
...
+    })
   }
   const onSubmit = (event: React.FormEvent) => {
     event.preventDefault()
+    const [brand, precinctId, ballotStyleId] = ballotCode.split('.')
```
benadida: my preference, in these situations where a critical security feature has a few lines of logic, is to make that a function with a clear name, even if it's only used once. Otherwise, it's hard to tell what's security critical and what is not.

src/App.tsx
beausmith: fixed

src/pages/ActivationPage.tsx
beausmith: fixed

src/components/Ballot.tsx
beausmith: I'm not sure I understand what you mean. Isn't this what the code is doing…
```
{contests.length ? '/contests/0' : '/'}
       ^                 ^          ^
  conditional          action     action
```

benadida: I'm not a fan of having the conditional be on the target of the redirect, when the outcome is so different, but it's also a nit. I should have been clearer.

src/components/Ballot.tsx
beausmith: Pushing an update to make it more readable. 

(Though, as we really have no direct need for a config option to enable/disable activation, I'd prefer to just remove the config options 😄)

src/pages/ActivationPage.tsx
beausmith: Ok, done did dis ding.

beausmith:
@benadida - Pls review latest commit. If looks good. Let's squash and merge this branch. Then resolve to make smaller changes in the future!

benadida:
@beausmith looks good to me!

PR #286, opened 2019-05-02
Add card functions

benadida:
add backend that talks to the smart card and frontend that queries the backend.


src/pages/ActivationPage.tsx
```diff
@@ -63,6 +66,23 @@ const StartPage = (props: RouteComponentProps) => {
     event.target.focus()
   }
 
+  // TODO: Mock this timer
+  /* istanbul ignore next */
```
beausmith: We previously agreed that anytime we were to use `istanbul ignore next` that we would document why and/or where the code is tested. eg: `/* istanbul ignore next: Tested by Cypress */`

src/pages/ActivationPage.tsx
```diff
@@ -63,6 +66,23 @@ const StartPage = (props: RouteComponentProps) => {
     event.target.focus()
   }
 
+  // TODO: Mock this timer
```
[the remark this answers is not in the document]
benadida: good callout!

src/pages/ActivationPage.test.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import { Route } from 'react-router-dom'
+
+import { render } from '../../test/testUtils'
+
+import ActivationPage from './ActivationPage'
+
+// https://www.npmjs.com/package/jest-fetch-mock
```
beausmith: nit: delete comment.

src/pages/ActivationPage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+import fetch, { FetchMock, GlobalWithFetchMock } from 'jest-fetch-mock'
+const customGlobal: GlobalWithFetchMock = global as GlobalWithFetchMock
```
beausmith: This line can simply be: `const customGlobal = global as GlobalWithFetchMock`

src/pages/ActivationPage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+customGlobal.fetch = fetch as FetchMock
+customGlobal.fetchMock = customGlobal.fetch as FetchMock
```
beausmith: This line can simply be: `customGlobal.fetchMock = customGlobal.fetch`

src/pages/ActivationPage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    [JSON.stringify({}), { status: 200 }],
+    [JSON.stringify({ card: 'VX.precinct-21.5R' }), { status: 200 }],
+    ['', { status: 500 }]
+  )
+
+  render(<Route path="/" component={ActivationPage} />, {
+    route: '/',
+  })
+
+  expect(setInterval).toHaveBeenCalledTimes(1)
```
beausmith: Let's avoid naked globals. Please change to `expect(window.setInterval).toHaveBeenCalledTimes(1)`

src/pages/ActivationPage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  expect(fetchMock.mock.calls.length).toEqual(3)
+  expect(fetchMock.mock.calls).toEqual([
+    ['/card/read'],
+    ['/card/read'],
+    ['/card/read'],
+  ])
+
+  // TODO: not clear why this one is failing
+  // expect(clearInterval).toHaveBeenCalledTimes(2)
```
beausmith: issue link?

src/pages/ActivationPage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    route: '/',
+  })
+
+  expect(setInterval).toHaveBeenCalledTimes(1)
+
+  jest.advanceTimersByTime(1000)
+
+  jest.advanceTimersByTime(1000)
+
+  // TODO: check that setActivationCode has been called
```
beausmith: Fix: `decodeActivationCode()` calls `setBallot()` which calls `activateBallot()` which is a context method which you can place in the `render()` method's second argument… and check to see if it has been called.

src/pages/ActivationPage.test.tsx
beausmith: Seem like this should work. Is `clearInterval` defined? Did you try `window.clearInterval`?

src/pages/ActivationPage.test.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
```
beausmith: Thank you for creating this file! Not sure how it was overlooked before. 🤦‍♂

src/pages/ActivationPage.tsx
```diff
@@ -77,6 +79,23 @@ const StartPage = (props: RouteComponentProps) => {
...
+      .then(resultJSON => {
+        const cardState = resultJSON.card
+        if (!!cardState) {
+          clearInterval(checkCard)
+          setActivationCode(cardState)
+          decodeActivationCode()
+        }
+      })
+      .catch(() => {
+        // if it's an error, aggressively assume there's no backend and stop hammering
```
beausmith: ❤️

src/pages/ActivationPage.tsx
```diff
@@ -77,6 +79,23 @@ const StartPage = (props: RouteComponentProps) => {
...
+        clearInterval(checkCard)
```
beausmith: Let's avoid naked globals. Please change to `window.clearInterval(checkCard)`

src/pages/ActivationPage.tsx
```diff
@@ -77,6 +79,23 @@ const StartPage = (props: RouteComponentProps) => {
     event.target.focus()
   }
 
+  checkCard = window.setInterval(() => {
+    fetch('/card/read')
+      .then(result => result.json())
+      .then(resultJSON => {
+        const cardState = resultJSON.card
+        if (!!cardState) {
+          clearInterval(checkCard)
```
beausmith: Let's avoid naked globals. Please change to `window.clearInterval(checkCard)`

src/pages/ActivationPage.test.tsx
benadida: the comment was to explain the approach, which isn't obvious, I think.

src/pages/ActivationPage.test.tsx
benadida: 👍 will try that

src/pages/ActivationPage.test.tsx
benadida: yep, doesn't work either. I'll update the commented call to make sure we don't keep the bad global, but I'm confused as to why this fails.

PR #291, opened 2019-05-03
Informed ignores

beausmith:
- All `istanbul ignore next` comments have note and/or issue.
- Replaced a couple ignore comments with tests!


src/AppActivationCodes.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const activationCodeInput = getByTestId('activation-code') as HTMLInputElement
+
+  // Set invalid activation code and do not submit form
+  fireEvent.change(getByTestId('activation-code'), {
+    target: { value: 'VX.precinct-21.5R' },
+  })
+  expect(activationCodeInput.value).toBe('VX.precinct-21.5R')
+
+  // Activation code input value is cleared.
+  jest.runAllTimers()
```
benadida: in case we later have `setInterval` used in other functions (ahem, I think I know of one), I recommend `runOnlyPendingTimers` as the default approach to testing.

src/setupTests.tsx
```diff
@@ -2,3 +2,22 @@
...
+
+// Supress "test was not wrapped in act()" Warning
+// Remove when upgrading to react-dom@16.9.0
+// Bug: https://github.com/facebook/react/pull/14853
+// Fix: https://github.com/testing-library/react-testing-library/issues/281#issuecomment-480349256
+/* tslint:disable:no-console no-any */
+const originalError = console.error
+beforeAll(() => {
+  console.error = (...args: any[]) => {
+    if (/Warning.*not wrapped in act/.test(args[0])) {
```
benadida: wow, that's some heavy duty suppression! Are we sure we care about suppressing warnings, given risk of overriding `console.error`?

beausmith: This is only in testing, and only filters this one error. 

Alternately, we can remove this and just ignore the errors as I believe they are non-blocking.

PR #297, opened 2019-05-03
Added data in QR code

benadida:
Adding the ballot data in QR code for redundancy.

This eventually needs to be compressed, right now it's inefficient and creates a somewhat dense QR code. Good for now, Will optimize in a further PR.


test/testUtils.tsx
```diff
@@ -17,11 +17,11 @@ export function render(
   {
     route = '/',
     activateBallot = jest.fn(),
-    ballotStyleId = '',
+    ballotStyleId = '5R',
     contests = electionSample.contests as Contests,
     election = electionSample,
     history = createMemoryHistory({ initialEntries: [route] }),
-    precinctId = '',
+    precinctId = 'precinct-21',
```
beausmith: Why do `ballotStyleId ` and `precinctId` have default values? They don't have default values when a voter walks up to the BMD. 

Mentioned this in previous PR:
https://github.com/votingworks/bmd/pull/235#discussion_r279603186

src/pages/__snapshots__/PrintPage.test.tsx.snap
```diff
@@ -531,7 +531,7 @@ exports[`renders PrintPage with votes 1`] = `
                   Precinct Number
                 </div>
                 <strong>
-                  21
+                  precinct-21
```
beausmith: https://votingworks.slack.com/archives/CEL6D3GAD/p1556658341003100
> No good reason why the mismatch, we should standardize, but we should pick the shorter approach because more characters quickly blows up the QR code.
> So VX.21.5R  

Are we going to standardize these ids?

benadida: Standardize only insofar as the IDs are going to come from the state systems and we don't control them. So we should be internally consistent with our tests but we can't be certain what other systems will do.

PR #321, opened 2019-05-06
Fix review page for numeric contest ids

benadida:
This was breaking for a different sample file where `contest.id` was only numeric, which makes for a disallowed DOM node `id`. Added a prefix and a test to make sure we don't later regress.

### DIFF
--- a/cypress/tests/scroll-to-contest.js
+++ b/cypress/tests/scroll-to-contest.js
@@ -13,9 +13,9 @@ describe('Review Page', () => {
     cy.wrap(clickThoughPages).each(() => {
       cy.contains('Next').click()
     })
-    cy.get('#county-commissioners').click()
+    cy.get('#contest-county-commissioners').click()
     cy.contains('Review Ballot').click()
-    cy.get('#county-commissioners').should('be.visible')
-    cy.focused().should('have.attr', 'id', 'county-commissioners')
+    cy.get('#contest-county-commissioners').should('be.visible')
+    cy.focused().should('have.attr', 'id', 'contest-county-commissioners')
   })
 })
--- a/src/pages/ReviewPage.test.tsx
+++ b/src/pages/ReviewPage.test.tsx
@@ -14,6 +14,8 @@ const contest1candidate0 = contest1.candidates[0]
 
 import ReviewPage from './ReviewPage'
 
+const idValueStartsWithNumberRegex = new RegExp('id="[0-9]')
+
 it(`renders ReviewPage without votes`, () => {
   const { container } = render(
     <Route path="/review" component={ReviewPage} />,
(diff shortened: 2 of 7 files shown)


src/pages/ReviewPage.test.tsx
```diff
@@ -14,6 +14,9 @@ const contest1candidate0 = contest1.candidates[0]
 
 import ReviewPage from './ReviewPage'
 
+// a regexp to test for DOM IDs that start with a digit, which is a no-no.
+const badDOMIdRegexp = new RegExp('id="[0-9]')
```
beausmith: (nit-ish… but I think this is important to balance the volume of comments in the code.)

Naming is hard… but important:
- If the variable name is updated to add clarity such that comment is unnecessary, then we don't have to maintain comment in addition to variable.
- This line simply defines a pattern, there is no comparison here or anything "bad", thus the comment and var name do not describe what is happening.

I suggest moving line 18 down to where it is used and renaming to:

```
const idValueStartsWithNumberRegex = new RegExp('id="[0-9]')
expect(idValueStartsWithNumberRegex.test(container.outerHTML)).toBeFalsy()
```

Note: this all may be moot for this instance given comment below.

src/pages/ReviewPage.test.tsx
```diff
@@ -35,5 +38,7 @@ it(`renders ReviewPage with votes`, () => {
       },
     }
   )
+
+  expect(badDOMIdRegexp.test(container.outerHTML)).toBeFalsy()
```
beausmith: I think this test is unnecessary because:
- this test will never be run in production (and thus never catch a production error).
- this test doesn't cover all app code where numeric `id` values may exist.
- we have prefixed the id value in the template… so there is no chance of this occurring.
- it is not testing anything which a user will encounter.

Do you agree?

cypress/tests/scroll-to-contest.js
```diff
@@ -13,9 +13,9 @@ describe('Review Page', () => {
     cy.wrap(clickThoughPages).each(() => {
       cy.contains('Next').click()
     })
-    cy.get('#county-commissioners').click()
+    cy.get('#contest-county-commissioners').click()
```
beausmith: Nothing new introduced here, but a note for future test writing… we really need to be selecting things based upon descriptions of things which the voter can find and click on… eg… not element ids.

src/pages/ReviewPage.test.tsx
benadida: oh excellent point, but I think I want to resolve it a different way: let's introduce some contest IDs that do start with a digit in `electionSample.json`. Then this test *will* be useful at ensuring that our code protects against numerical contest IDs, which we will see in config files produced by other scripts/code sometimes not our own.

What do you think?

src/pages/ReviewPage.test.tsx
benadida: I like the more semantic variable name. Good call.

src/pages/ReviewPage.test.tsx
benadida: wait, actually, I already did this :) My past self is smarter than my today self. 

I changed `measure-102` to `102` for the sole purpose of making sure this test would catch any code change that fails to protect DOM node IDs.

PR #341, opened 2019-05-11
 Config and run stylelint correctly.

beausmith:
Figured out how to use stylelint for both css and css-in-js files.

### DIFF
--- a/cypress/tests/scroll-to-contest.js
+++ b/cypress/tests/scroll-to-contest.js
@@ -1,9 +1,4 @@
-import electionSample from '../../src/data/electionSample'
-
-const clickThoughPages = [
-  ...electionSample.contests.slice(0, 20),
-  { id: 'pre-review-page' },
-]
+const clickThoughPages = new Array(21) // number of contests + pre-review screen for activation code 'VX.23.12'
 
 describe('Review Page', () => {
   it('When navigating from contest, scroll to contest and place focus on contest.', () => {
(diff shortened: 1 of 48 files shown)


cypress/tests/scroll-to-contest.js
```diff
@@ -1,9 +1,4 @@
-import electionSample from '../../src/data/electionSample'
-
-const clickThoughPages = [
-  ...electionSample.contests.slice(0, 20),
-  { id: 'pre-review-page' },
-]
+const clickThoughPages = new Array(21) // number of contests + pre-review screen for activation code 'VX.23.12'
```
benadida: sneaking in an unrelated change I see??? ;) It's fine.

src/pages/PrintPage.tsx
```diff
@@ -214,7 +214,8 @@ class SummaryPage extends React.Component<RouteComponentProps, State> {
               <Header>
                 <div
                   className="seal"
-                  dangerouslySetInnerHTML={{ __html: seal }}
+                  // TODO: Sanitize the SVG content: https://github.com/votingworks/bmd/issues/99
```
benadida: 👍

cypress/tests/scroll-to-contest.js
beausmith: No sneaky business. 😄 Seems unrelated, but I was getting an ESLint error saying there was an issue importing the `electionSample.json` file (even though Cypress runs fine). While I was troubleshooting it, I realized that since ballot style support was added, this test no longer relies upon the data in the file to determine the length of contests… so why not just manually specify the length another way and remove the import?

PR #344, opened 2019-05-11
added proxying for development with yarn start and the backend server

benadida:
Using `package.json` proxying, development is now way easier with both front-end and backend at the same time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "@types/react-router-dom": "^4.3.3",
     "@types/styled-components": "^4.1.15",
     "history": "^4.9.0",
+    "http-proxy-middleware": "^0.19.1",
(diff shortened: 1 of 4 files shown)


benadida:
grumble, I think there's an issue with `start-server-and-test` where it doesn't send an `Accept` header and so the proxy kicks in, and boom. Working on it...

benadida:
@beausmith I've cleaned up the comments to clarify that, as best as I can tell:

* `setupProxy` does not support languages other than native Node
* `setupProxy` does not support ES6 style imports

As a result, I don't see a way to do the items you suggest. Even the `exclude` is required to make this work.

I've cleaned up the comment to try to make this limitation clearer.

PR #345, opened 2019-05-11
add architecture documentation

benadida:
@beausmith this is a documentation PR for the express purpose of discussing my proposal to structure the server in small components. The BMD one would likely only have one server component to start, smart-card reading/writing, but over time may add something like "print in the background so there's no flashing popup." Scanner and EMS will likely need more backend components, e.g. to perform the scanning itself, or to produce the results file.

My goal is to reuse code, so the smart-card code doesn't need to be duplicated across BMD, BAS, EMS, etc... while also not having unnecessary functionality in the BMD backend, for example.

What this means, if we agree on this approach, is that the BMD repository's server component will shrink to be just a simple web proxy to the backend server components it needs. Those server components will each be implemented in their own repo.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,37 @@
+# Architecture
+
+This repository, BMD, is a front-end React application that can do just about
+everything in front-end web. However, some features are beyond the web
+environment:
+
+- reading / writing smart cards
+- printing without a flashing dialog (darn Chrome Kiosk mode)
+- possibly more in the future
+
+## The "Big" Idea
+
+To make these features available to the front-end, we serve the React static
+HTML+CSS+JS from a localhost web server, and that localhost web server provides
+URL endpoints that implement the missing features using code running directly on
+the machine.
+
+## DRY & Minimizing Features
+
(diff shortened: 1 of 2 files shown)


docs/architecture.md
```diff
@@ -0,0 +1,37 @@
...
+everything in front-end web. However, some features are beyond the web
+environment:
+
+- reading / writing smart cards
+- printing without a flashing dialog (darn Chrome Kiosk mode)
+- possibly more in the future
+
+## The "Big" Idea
+
+To make these features available to the front-end, we server the React static
```
beausmith: typo `server` --> `serve`

docs/architecture.md
```diff
@@ -0,0 +1,37 @@
...
+HTML+CSS+JS from a localhost web server, and that localhost web server provides
+URL endpoints that implement the missing features using code running directly on
+the machine.
+
+## DRY & Minimizing Features
+
+The ability to read/write smart cards is important for the BMD, the Ballot
+Activation Station (BAS), the scanner, and the Election Mnaagement System (EMS).
```
beausmith: typo

PR #363, opened 2019-05-15
Card polling in app

benadida:
Move card polling to the App level and fix testing.


src/AppCard.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    ['/card/read'],
+  ])
+
+  jest.useRealTimers()
+})
+
+it(`CardData processing processes card data properly`, () => {
+  const div = document.createElement('div')
+  // @ts-ignore - App expects ReactRouter props, but are unnecessary for this test.
+  const app = (ReactDOM.render(<App />, div) as unknown) as App
```
beausmith: I really wanted to find a better way to do this, but was unable to do it within my self-imposed time box. 😐

src/AppCard.test.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react'
```
beausmith: This tests in this file work, but shouldn't we be testing from the perspective of the user and not be reaching into the internals which the user cannot directly react? 

Do you agree with this: https://testing-library.com/docs/guiding-principles

src/AppCard.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+        present: true,
+        shortValue: JSON.stringify({
+          t: 'activation',
+          pr: election.precincts[0].id,
+          bs: election.ballotStyles[0].id,
+        }),
+      }),
+      { status: 200 },
+    ],
+    ['', { status: 500 }]
```
beausmith: Should this status be `200`? 

Where can I find more info on these status values?

src/AppCard.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  window.location.href = '/'
+})
+
+it(`App fetches the card data every 1 second`, () => {
+  fetchMock.resetMocks()
+  jest.useFakeTimers()
+
+  fetchMock.mockResponses(
+    [JSON.stringify({}), { status: 200 }],
+    // This response covers the App functionality for processing card data.
```
beausmith: I added this comment because in this test file the data contained in the following response is not confirmed to have been received later in this same test… however it does satisfy code coverage.

src/App.tsx
```diff
@@ -64,8 +67,33 @@ const initialState = {
   votes: {},
 }
 
-class App extends React.Component<RouteComponentProps, State> {
+export class App extends React.Component<RouteComponentProps, State> {
   public state: State = initialState
+  public checkCard: number = 0
```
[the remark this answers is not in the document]
benadida: I like both of those points, will make the change.

src/App.tsx
```diff
@@ -117,7 +161,7 @@ class App extends React.Component<RouteComponentProps, State> {
     return voterData ? JSON.parse(voterData) : {}
   }
 
-  public setBalotActivation = (data: {
+  public setBallotActivation = (data: {
```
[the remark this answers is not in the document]
benadida: `stBltCtvtn`

who needs vowels, really?

src/config/types.ts
```diff
@@ -128,4 +128,19 @@ export interface BallotContextInterface {
   userSettings: UserSettings
 }
 
+// Smart Card Content
+export type CardDataTypes = 'activation' | 'admin'
```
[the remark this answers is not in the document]
benadida: love this point, will change accordingly.

src/AppCard.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const election = electionSample as Election
+
+beforeEach(() => {
+  window.localStorage.clear()
+  window.location.href = '/'
+  fetchMock.resetMocks()
+})
+
+it(`App fetches the card data every 1 second`, async () => {
```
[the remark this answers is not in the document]
benadida: 👍

src/AppCard.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+import ReactDOM from 'react-dom'
+import { render } from 'react-testing-library'
+
+import electionSample from './data/electionSample.json'
+
+import Root, { App } from './App'
+import {
+  ActivationCardData,
+  AdminCardData,
+  CardData,
```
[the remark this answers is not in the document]
benadida: I did not somehow, and I didn't skip commit warnings. Weird.

src/AppCard.test.tsx
benadida: Yeah, I tried a lot of different things, too.

src/AppCard.test.tsx
benadida: No this is meant to test that on a 500 or server error of any kind, the client bails out and thus calls `clearInterval`

src/AppCard.test.tsx
benadida: I think those are good principles for testing UI, but in the world where you're testing network protocol code, I think unit tests on key functions of the code can be very useful.

benadida: and to be clear, functional and integration tests that show the overall effect are super useful. I don't meant to take away from them. It's just hard to have them cover every use case, especially the error cases, without making tests pretty unwieldy.

PR #372, opened 2019-05-20
Ballot tracker

benadida:
Initial pass at implementing ElectionGuard ballot tracker.

@beausmith the styling and layout are not my areas of expertise, I did my best to copy your style, please tell me how to fix this :)

Also this doesn't have tests yet cause I want to validate approach before I write them.

### DIFF
--- a/src/AppEndToEnd.test.tsx
+++ b/src/AppEndToEnd.test.tsx
@@ -8,6 +8,9 @@ import electionSample from './data/electionSample.json'
 import App, { mergeWithDefaults } from './App'
 import { CandidateContest, Election, YesNoContest } from './config/types'
 
+// turn off voterBallotTracker for these tests
+electionSample.voterBallotTracker = false
+
 const electionSampleAsString = JSON.stringify(
   mergeWithDefaults(electionSample as Election)
 )
--- a/src/pages/TrackerPage.tsx
+++ b/src/pages/TrackerPage.tsx
@@ -1,0 +1,287 @@
+import React from 'react'
+import { RouteComponentProps } from 'react-router-dom'
+import styled from 'styled-components'
+import QRCode from '../components/QRCode'
+import Seal from '../components/Seal'
+import { encryptAndGetTracker } from '../endToEnd'
+
+import ButtonBar from '../components/ButtonBar'
+import Main, { MainChild } from '../components/Main'
+import Prose from '../components/Prose'
+import BallotContext from '../contexts/ballotContext'
+
+const Header = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  height: 0.75in;
+  & h2 {
+    margin-bottom: 0;
+  }
+  & h3 {
+    margin-top: 0;
+  }
+  & > .ballot-header-content {
+    flex: 1;
+    margin: 0 1rem;
+    max-width: 100%;
+  }
+`
+const Content = styled.div`
+  flex: 1;
+  line-height: 1.2;
+`
+
+const Highlight = styled.div`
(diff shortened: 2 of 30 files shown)


beausmith:
Hmmmm… I had been thinking we'd implement a much simpler approach. 

Rather than make this a second "web" page, I was planning to optionally render this new "component" as a second "printed" page which is rendered in `PrintPage.tsx`. This way both pages print at the same time.

Perhaps best if you simply add code to create and output a tracker code to `PrintPage.tsx` for now?

benadida:
I guess I'm not sure that's simpler? Feels like there are potential edge conditions to consider if the ballot itself goes over a single page, and then making sure the tracking number is printed separately.

Consider too that, at some future point, we may want to print the tracking number from a different printer or different sheet feeder.

Overall, my feeling is that it will be simpler to make the tracker a separate print job.

beausmith:
Let's discuss. 

More context:
- There are print styles to enforce that an element start on a new piece of paper.
- Asking the voter to print again, or having this be printed from another printer, etc are not flows we've discussed at all.

benadida:
After in-person discussion, some clarifications:
- there is a chance this eventually is a separate printer, let's account for that.
- definitely don't need user interaction to print the second sheet.

Will fix tests! @beausmith do let me know if there is anything else you want to comment on in this PR. But maybe after tests.

src/AppEndToEnd.test.tsx
```diff
@@ -6,6 +6,9 @@ import electionSample from './data/electionSample.json'
 import App, { electionKey, mergeWithDefaults } from './App'
 import { CandidateContest, Election, YesNoContest } from './config/types'
 
+// turn off voterBallotTracker for these tests
+electionSample.voterBallotTracker = false
```
beausmith: Sorry I didn't catch this sooner, but it seems that `voterBallotTracker` is an option which can be configured on/off and should be placed into [electionDefaults.json](https://github.com/votingworks/bmd/blob/master/src/data/electionDefaults.json) with  default of `false`. Let's make this change.

src/pages/TrackerPage.tsx
```diff
@@ -0,0 +1,143 @@
...
+                  </p>
+                </Prose>
+              </Header>
+              <Content>
+                <Highlight>
+                  <Callout>&#8594;</Callout>
+                  <b>
+                    Use the unique tracking ID below to track your individual
+                    ballot and see it’s been recorded.
+                  </b>{' '}
```
beausmith: nit: `<strong>` is accessible, `<b>` is only presentational.

src/pages/TrackerPage.tsx
```diff
@@ -0,0 +1,143 @@
...
+                  Although the contents of your ballot will always remain
+                  secret, you can verify that the vote recorded was at the time
+                  and place that you voted.
+                </Highlight>
+                <p />
```
beausmith: nit: if the `Highlight` component was a `p` element you wouldn't need to use this `<p />`

The `<Prose>` component is for styling hierarchical text elements. If all it's children are h1, h2, p, ul, table, etc. they will be nicely spaced.

Don't bother changing this. I'll give it a pass later.

beausmith:
Note: I added one small types-related simplification commit to this branch.

src/AppEndToEnd.test.tsx
benadida: I'm not sure that's right, though tell me what you think: it's meant to be something that's found in `election.json`, not something that is later configurable. `election.json` is the final word. This test line is meant to do the end-to-end test without ballot tracking without also having to duplicate the sample file.

PR #382, opened 2019-05-22
updates for new smart card formats

benadida:
tested on smartcard hardware, works!

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -112,7 +112,7 @@ export interface BallotContextInterface {
 }
 
 // Smart Card Content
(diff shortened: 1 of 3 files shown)


src/config/types.ts
```diff
@@ -112,7 +112,7 @@ export interface BallotContextInterface {
 }
 
 // Smart Card Content
-export type CardDataTypes = 'voter' | 'admin'
+export type CardDataTypes = 'voter' | 'pollworker' | 'clerk'
```
beausmith: nit for later: "CardData" refers to short value only, but one could easily assume that it is all the card data.

PR #383, opened 2019-05-23
Full use of smart cards

benadida:


src/App.tsx
```diff
@@ -118,22 +124,34 @@ export class App extends React.Component<RouteComponentProps, State> {
         ) {
           this.setState({ loadingElection: true })
           this.fetchElection().then(election => {
-            this.setElection(JSON.parse(election.longValue))
+            // setTimeout to prevent tests from going into infinite loops
+            window.setTimeout(() => {
```
benadida: this one reduces the number of warnings, but doesn't fully make them go away.

PR #387, opened 2019-05-23
add timeout to resetballot cause recursion

benadida:
apparently those warnings during tests can matter... this removes one of them and fixes an actual bug in the last screen. I'll work next to remove the remaining warning.

### DIFF
--- a/src/AppEndToEnd.test.tsx
+++ b/src/AppEndToEnd.test.tsx
@@ -208,6 +208,7 @@ it(`basic end-to-end flow`, async () => {
   })
 
   // Review and Cast Instructions
+  await sleep(100)
   getByText('Verify and Cast Your Ballot')
 
   // ===========================================================================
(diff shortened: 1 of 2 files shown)


src/AppEndToEnd.test.tsx
```diff
@@ -208,6 +208,7 @@ it(`basic end-to-end flow`, async () => {
   })
 
   // Review and Cast Instructions
+  await sleep(100)
```
beausmith: Please add a comment here to indicate **why** this line exists.

src/pages/PrintPage.tsx
```diff
@@ -163,7 +163,9 @@ class SummaryPage extends React.Component<RouteComponentProps, State> {
     window.removeEventListener('afterprint', this.resetBallot)
   }
   public resetBallot = () => {
-    this.context.resetBallot('/cast')
+    window.setTimeout(() => {
```
beausmith: Please add a comment here to indicate **why** this timeout exists.

PR #389, opened 2019-05-24
random serial number, works with write-ins

benadida:
a real random serial number, and encoding of choices supports write-ins.


src/utils/random.ts
```diff
@@ -0,0 +1,8 @@
+export default function randomBase64(numBytes: number) {
+  // get random values to fill an array of prescribed length, b64 encode it, and strip the '=' at the end.
+  return btoa(
```
beausmith:
```suggestion
  return window.btoa(
```

src/utils/random.ts
```diff
@@ -0,0 +1,8 @@
...
+    String.fromCharCode.apply(undefined, (window.crypto.getRandomValues(
+      new Uint8ClampedArray(numBytes)
+    ) as unknown) as number[])
+  ).slice(0, -2)
```
beausmith:
```suggestion
  ).replace(/=.*$/, '')
```

src/votecoding.ts
```diff
@@ -0,0 +1,34 @@
...
+        } else {
+          return '0'
+        }
+      }
+
+      const candidateIDs = (contest as CandidateContest).candidates.map(
+        (c: Candidate) => c.id
+      )
+      return (contestVote as CandidateVote)
+        .map(c => (c.isWriteIn ? 'W' : candidateIDs.indexOf(c.id)))
```
beausmith: Why not encode the write-in value submitted by the voter such that the ballot QR Code and the write-in value printed on the ballot?

src/votecoding.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (!contestVote) {
+        return ''
+      }
+
+      if (contest.type === 'yesno') {
+        if (contestVote === 'yes') {
+          return '1'
+        } else {
+          return '0'
+        }
```
beausmith: Suggestion: collapse 5 lines into 1?
```js
return contestVote === 'yes' ? '1' : '0'
```

src/utils/random.ts
beausmith: Or…

```suggestion
  ).replace(/==$/, '')
```

src/pages/PrintPage.tsx
```diff
@@ -3,6 +3,8 @@ import { RouteComponentProps } from 'react-router-dom'
 import styled from 'styled-components'
 import QRCode from '../components/QRCode'
 import { findPartyById } from '../utils/find'
+import randomBase64 from '../utils/random'
+import encodeVotes from '../votecoding'
```
beausmith: nit: There was an ESLint rule to check for this (I'll have to look into what happened), but rule enforces that default export should have same name as file. Suggestion:
```js
import random from '../utils/random'
import encodeVotes from '../encodeVotes'
```

src/utils/random.ts
benadida: 👍

src/utils/random.ts
benadida: cool, sounds good.

src/votecoding.ts
benadida: because
1/ space is limited in QR code, and a few write-ins will shrink the pixels and make it harder to decode.
2/ then we have to worry about escaping characters properly so our parsing doesn't get in trouble (super tricky to do fully well without going full JSON)

src/votecoding.ts
benadida: 👍

src/pages/PrintPage.tsx
benadida: `encodeVotes` sure, but I'd like to keep the `randomBase64` because I think that's more descriptive.

src/utils/random.ts
benadida: doing it slightly differently cause it could be one or two `=`, but same idea.

src/votecoding.ts
benadida: also, write-ins tend to be a different process, not counted unless they go above a certain amount.

PR #392, opened 2019-05-24
If this is a demo election (no module-smartcards), things should still work

benadida:


src/AppEndToEndWithoutModuleSmartcards.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  window.location.href = '/'
+})
+
+async function sleep(milliseconds: number) {
+  return new Promise(resolve => {
+    window.setTimeout(resolve, milliseconds)
+  })
+}
+
+it(`quick end-to-end flow`, async () => {
```
beausmith: nit: more descriptive test title please.

src/AppEndToEndWithoutModuleSmartcards.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const { getByText, getByTestId, queryByText } = render(<App />)
+
+  fireEvent.click(getByText('Load Sample Election File'))
+
+  // wait long enough to get the /card/read to fail and flip the demo bit
+  await sleep(250)
+
+  getByText('Scan Your Activation Code')
+  fireEvent.click(getByTestId('qrContainer'))
```
beausmith: How do you image we should get past the activation screen once we replace QR Code on the activation page with some type of card-reader graphic. :thinking:

benadida: great q.

I think we'll want to identify a demo mode, probably in the `election.json`, where you can just tap the smart card icon. But not in a real election of course.

src/AppEndToEndWithoutModuleSmartcards.test.tsx
benadida: 👍

PR #399, opened 2019-05-27
Makefile and prod web server

benadida:
To run this in production, a simple Makefile and node server setup that reuses `setupProxy.js`. All JavaScript for now because of this reuse of `setupProxy.js`. We can consider upgrading this all to Typescript when `setupProxy` supports it.


prodserver/index.js
```diff
@@ -0,0 +1,17 @@
+// This file runs a simple production web server of the React production build
+// it reuses setupProxy.js for consistent use of backend modules in dev and prod.
+// This is plain JS rather than our TS setup in the React app because of that.
+//
+/* eslint-disable */
```
beausmith:
```suggestion
```

prodserver/index.js
```diff
@@ -0,0 +1,17 @@
...
+/* istanbul ignore file */
+
+const express = require('express')
```
beausmith:
```suggestion
import express from 'express'
```

prodserver/index.js
```diff
@@ -0,0 +1,17 @@
...
+const proxy = require('../src/setupProxy')
```
beausmith:
```suggestion
import proxy from '../src/setupProxy'
```

prodserver/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+const app = express()
+const port = 3000
+
+proxy(app)
+app.use('/', express.static('../build'))
+
+app.listen(port, () => console.log(`BMD listening on port ${port}!`))
```
beausmith:
```suggestion
app.listen(port, () => console.log(`BMD listening on port ${port}!`)) // eslint-disable-line
```

benadida:
@beausmith I believe I tried that approach you suggest, and it only works because `tsconfig.json` only includes `src`. The moment you try to have it include `prodserver`, it starts complaining about importing a file that uses a `require`, and I couldn't find a way around that. So, if you want to go this route, I can, but I think we're kinda fooling ourselves and we're not actually making `tsc` happy.

beausmith:
Cool. Understood. I'm only trying to apply ESLint here as I don't see a reason not to lint this file. This suggesting has nothing to do with TypeScript.

benadida:
after discussion in person with @beausmith, given difficulty of having node use modern import syntax, going to stick with what we have.

PR #405, opened 2019-05-27
Interstitials

### LINKED ISSUE #263: Mocks for interstitial design and content
As mentioned in today's UX team meeting, here are the most current designs and content for the 4 interstitial pages. They may need further adjustments as we go forward with testing but should make a solid starting point. For example, debatable if "Cast" is the right word for the final interstitial.

Happy to discuss details further. 

[Oxide-F&F-Microsoft-DDT-Interstitials-01.pdf](https://github.com/votingworks/bmd/files/3113450/Oxide-F.F-Microsoft-DDT-Interstitials-01.pdf)

### LINKED ISSUE #393: Interstitials

beausmith:
## Related Issue(s)

Resolves #393 
Resolves #263 

## Description

- Add Instructions page.
- Create Breadcrumbs component.

## Screenshots/Screencasts

image

image

image

image


src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,101 @@
...
+  width: calc(100% + 2rem);
+  text-align: center;
+  color: ${({ current }) => (current ? undefined : '#c7d7df')};
+  font-weight: ${({ current }) => (current ? '600' : undefined)};
+`
+
+const Breadcrumbs = ({ step }: { step: number }) => (
+  <BreadcrumbContainer>
+    <StepContainer>
+      {['Mark', 'Review', 'Print', 'Cast'].map((label, i) => {
```
benadida: nit, not a blocker, possible future enhancement: you've made a nice component that could be reused for any wizard type thing, but the steps are hard-wired. Feels like those should be props or children elements to `Breadcrumbs`, especially since the instructions that match the step are provided outside of this component, so there's a bit of an abstraction mismatch.

beausmith: Yeah, thought about this. The component only supports 4 steps, so I'm kinda enforcing this by hard coding the steps in. Agreed on the abstraction though. I didn't want to spend any more time making this support more than 4 steps. I guess I could have added a generic check in the component to error if more than four steps.

beausmith: Yeah, thought about this. The component only supports 4 steps, so I'm kinda enforcing this by hard coding the steps in. Agreed on the abstraction though. I didn't want to spend any more time making this support more than 4 steps. I guess I could have added a generic check in the component to error if more than four steps.

PR #411, opened 2019-05-29
a number of small changes to make the audio ballot work

benadida:
Moving away from strict web accessibility so it is very voting-machine-function specific.

### DIFF
--- a/src/components/CandidateContest.tsx
+++ b/src/components/CandidateContest.tsx
@@ -181,6 +181,7 @@ const Choice = styled('label')<{ isSelected: boolean }>`
 `
 const ChoiceInput = styled.input.attrs({
   type: 'checkbox',
+  role: 'option',
 })`
   margin-right: 0.5rem;
 `
@@ -418,12 +419,16 @@ class CandidateContest extends React.Component<Props, State> {
       <React.Fragment>
         <Main noOverflow noPadding>
           <ContentHeader id="contest-header">
-            <Prose>
-              <h1 aria-label={`${contest.section}, ${contest.title}.`}>
+            <Prose id="audiofocus">
+              <h1 aria-label={`${contest.title}.`}>
                 <ContestSection>{contest.section}</ContestSection>
                 {contest.title}
               </h1>
-              <p>
+              <p
+                aria-label={`Vote for ${contest.seats}. You have selected ${
+                  vote.length
+                }. Use the down arrow to hear your options. Use the right arrow to move to the next contest.`}
+              >
                 <strong>Vote for {contest.seats}.</strong> You have selected{' '}
                 {vote.length}.
               </p>
@@ -438,7 +443,7 @@ class CandidateContest extends React.Component<Props, State> {
               onScroll={this.updateContestChoicesScrollStates}
             >
               <ScrollableContentWrapper isScrollable={isScrollable}>
-                <ChoicesGrid role="group" aria-labelledby="contest-header">
+                <ChoicesGrid>
                   {contest.candidates.map(candidate => {
                     const isChecked = !!this.findCandidateById(
                       vote,
@@ -459,6 +464,9 @@ class CandidateContest extends React.Component<Props, State> {
                         htmlFor={candidate.id}
                         isSelected={isChecked}
                         onClick={handleDisabledClick}
+                        aria-label={`${candidate.name}, ${
+                          party ? party.name : ''
+                        }.`}
                       >
                         <ChoiceInput
                           id={candidate.id}
@@ -473,12 +481,7 @@ class CandidateContest extends React.Component<Props, State> {
                           className="visually-hidden"
                         />
                         <Prose>
-                          <Text
-                            wordBreak
-                            aria-label={`${candidate.name}, ${
-                              party ? party.name : ''
-                            }.`}
-                          >
+                          <Text wordBreak>
                             <strong>{candidate.name}</strong>
                             <br />
                             {party ? party.name : ''}
@@ -553,20 +556,27 @@ class CandidateContest extends React.Component<Props, State> {
         </Main>
         <Modal
           isOpen={!!attemptedOvervoteCandidate}
+          ariaLabel=""
           content={
             <Prose>
-              <Text>
+              <Text id="modalaudiofocus">
                 You may only select {contest.seats}{' '}
                 {contest.seats === 1 ? 'candidate' : 'candidates'} in this
                 contest. To vote for{' '}
                 {attemptedOvervoteCandidate && attemptedOvervoteCandidate.name},
                 you must first unselect selected{' '}
                 {contest.seats === 1 ? 'candidate' : 'candidates'}.
+                <span aria-label="Use the select button to continue." />
               </Text>
             </Prose>
           }
           actions={
-            <Button primary autoFocus onClick={this.closeAttemptedVoteAlert}>
+            <Button
+              primary
+              autoFocus
+              onClick={this.closeAttemptedVoteAlert}
+              aria-label="use the select button to continue."
+            >
               Okay
             </Button>
           }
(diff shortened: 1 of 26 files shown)


src/pages/CastBallotPage.tsx
```diff
@@ -55,8 +55,14 @@ const VoteInstructionsPage = () => {
             <Text narrow>
               If you find a mistake, ask a poll worker for help.
             </Text>
+            <div aria-label="Press the down arrow to continue, then" />
```
beausmith: then…? Does this make sense?

src/pages/InstructionsPage.tsx
```diff
@@ -15,13 +15,19 @@ const InstructionsPage = () => {
...
             <h1 aria-label="Mark your ballot.">Mark your ballot</h1>
             <Text narrow>{`This ballot has ${contests.length} contests.`}</Text>
             <p>
-              <LinkButton primary big to="/contests/">
+              <LinkButton
+                primary
+                big
+                to="/contests/"
+                id="next"
+                aria-label="Select Next to Start Voting."
```
beausmith: This seems different… that the other instructions. Perhaps…

```suggestion
                aria-label="Use the right arrow to Start Voting."
```

…or…

```suggestion
                aria-label="Press the select button to Start Voting."
```

src/pages/PreReviewPage.tsx
```diff
@@ -17,11 +17,17 @@ const SummaryPage = () => {
       <Main>
         <MainChild center>
           <Breadcrumbs step={2} />
-          <Prose textCenter>
+          <Prose textCenter id="audiofocus">
             <h1 aria-label="Review Your Selections.">Review Your Selections</h1>
             <Text narrow>Confirm and change any votes as necessary.</Text>
+            <div aria-label="Use the down arrow, then" />
```
beausmith: Wondering about this like the previous one. I think we should only use full sentences/phrases to keep things as translatable as possible. And also avoid editing one part without the other.

src/pages/PrintPage.tsx
```diff
@@ -203,15 +203,21 @@ class SummaryPage extends React.Component<RouteComponentProps, State> {
...
+            <Prose textCenter className="no-print" id="audiofocus">
+              <h1 aria-label="Print your official ballot.">
                 Print your official ballot
               </h1>
               <Text narrow>
                 If you have reviewed your selections and you are done voting,
                 you are ready to print your official ballot.
               </Text>
-              <Button primary big onClick={this.showConfirm}>
+              <span aria-label="First, press the down arrow, then" />
```
beausmith: Hmmm… I really need to test this myself. Wondering… can we use the `autoFocus` like in the modal to autoFocus the "Print Ballot" button?

src/pages/PrintPage.tsx
```diff
@@ -233,12 +239,14 @@ class SummaryPage extends React.Component<RouteComponentProps, State> {
         <Modal
           isOpen={this.state.showConfirmModal}
           centerContent
+          ariaLabel=""
```
beausmith: Why empty? Mistake?

src/pages/PrintPage.tsx
```diff
@@ -233,12 +239,14 @@ class SummaryPage extends React.Component<RouteComponentProps, State> {
...
           content={
-            <Prose>
+            <Prose id="modalaudiofocus">
               <Text center>
                 You may not make any changes after you print your ballot.
               </Text>
               <Text center>Do you want to print your ballot?</Text>
+              <span aria-label="Use the down arrow to continue." />
```
beausmith: I think this much safer/future-proof solution than ending aria-label's with "then".

src/pages/ReviewPage.tsx
```diff
@@ -326,7 +328,7 @@ class ReviewPage extends React.Component<RouteComponentProps, State> {
                     to={`/contests/${i}#review`}
                   >
                     <ContestProse compact>
-                      <h3 aria-label={`${contest.section}, ${contest.title},`}>
```
beausmith: Seeming like `contest.section` was removed on purpose. I feel bad that visually impaired voters are not getting the same content which sighted voters are getting.

src/pages/ReviewPage.tsx
```diff
@@ -326,7 +328,7 @@ class ReviewPage extends React.Component<RouteComponentProps, State> {
...
+                      <h3 aria-label={`${contest.title.replace(',', '')},`}>
```
beausmith: Too long of a pause? I'm worried we're gonna regret removing commas here.

src/pages/ReviewPage.tsx
```diff
@@ -344,8 +346,8 @@ class ReviewPage extends React.Component<RouteComponentProps, State> {
                         />
                       )}
                     </ContestProse>
-                    <ContestActions>
-                      <DecoyButton>Change</DecoyButton>
+                    <ContestActions aria-label="press enter to change your answer for this contest.">
+                      <DecoyButton aria-hidden="true">Change</DecoyButton>
```
beausmith: Why not place the label on the DecoyButton? Seems like that's where the "action" is? 😃

src/pages/StartPage.tsx
```diff
@@ -32,7 +32,12 @@ const StartPage = () => {
             Precinct {precinctId}
           </p>
           <p>
-            <LinkButton primary to="/instructions/">
+            <LinkButton
+              primary
+              to="/instructions/"
+              id="next"
+              aria-label="Select Next to Get Started."
```
beausmith: For consistency…

```suggestion
              aria-label="Press the right arrow to get started."
```

…or…

```suggestion
              aria-label="Press select to get started."
```

beausmith:
BTW, I'm very glad you took this on. I think it was great to have a fresh perspective take a pass.

src/components/CandidateContest.tsx
```diff
@@ -418,12 +419,16 @@ class CandidateContest extends React.Component<Props, State> {
       <React.Fragment>
         <Main noOverflow noPadding>
           <ContentHeader id="contest-header">
-            <Prose>
-              <h1 aria-label={`${contest.section}, ${contest.title}.`}>
+            <Prose id="audiofocus">
+              <h1 aria-label={`${contest.title}.`}>
```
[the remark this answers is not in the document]
benadida: Because it is super distracting and repetitive when the sound is on. It felt a lot better this way, and not much information lost.

src/components/CandidateContest.tsx
```diff
@@ -459,6 +464,9 @@ class CandidateContest extends React.Component<Props, State> {
                         htmlFor={candidate.id}
                         isSelected={isChecked}
                         onClick={handleDisabledClick}
+                        aria-label={`${candidate.name}, ${
+                          party ? party.name : ''
```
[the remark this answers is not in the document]
benadida: I don't think so. I have not seen that explicit denomination on the ballots we're targeting

src/pages/CastBallotPage.tsx
benadida: It works well because it reads the next button instructions too. It's the least disruptive way to make this work without restructuring the HTML

src/pages/PrintPage.tsx
benadida: Not a mistake, meant to override the default label that is inside Modal. Again, trying to disrupt as little as possible.

src/pages/PrintPage.tsx
benadida: It doesn't really work as well for the other flows. I promise, I've tried a lot of things :)

src/pages/ReviewPage.tsx
benadida: Because if I do that, it says "button" and I can't stop it from saying that :)

benadida: Because if I do that, it says "button" and I can't stop it from saying that :)

benadida:
after discussion with @beausmith, we're going to merge this since this is an improvement over what we have, but we are tracking these comments for improvement in #416

PR #440, opened 2019-05-31
Election data on activation screens

### LINKED ISSUE #394: Election data on waiting screens.

beausmith:
Fixes #394

### DIFF
--- a/src/pages/StartPage.tsx
+++ b/src/pages/StartPage.tsx
@@ -1,20 +1,11 @@
 import React, { useContext } from 'react'
-import styled from 'styled-components'
 
 import BallotContext from '../contexts/ballotContext'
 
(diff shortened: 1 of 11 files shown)


src/pages/StartPage.tsx
```diff
@@ -23,27 +14,22 @@ const StartPage = () => {
...
             <br />
             {county.name}, {state}
-            <br />
-            Ballot Style {ballotStyleId}
-            <br />
-            Precinct {precinctId}
           </p>
+          <hr />
+          <h2>
+            Precinct: {precinctId}
```
benadida: nit: I was wondering if we maybe should display the precinct name while we're at it. Kind of nicer. Not a blocker.

beausmith: Done.

PR #443, opened 2019-06-01
Accessible keyboard

beausmith:
If you want something done right, do it yourself.

@benadida - Review the first commit separately.


src/App.tsx
```diff
@@ -173,22 +173,33 @@ export class App extends React.Component<RouteComponentProps, State> {
...
+            !isClerkCardPresent
+          ) {
+            if (card.shortValue) {
+              const cardData = JSON.parse(card.shortValue) as CardData
+              this.processCardData({
+                cardData: cardData,
+                longValueExists: card.longValueExists,
+              })
+            } else {
+              this.setState({
```
benadida: I think this logic is wrong, because what happens when you remove a card? It needs to run this `setState` and I don't think it ever gets here. I'll check the actual running code.

benadida: yep, confirmed in running code, card removal is no longer processed.

Let's back out this commit to make sure we don't introduce other weird logic issues?

benadida:
after discussion with @beausmith, removing the card-state change commit, and merging.

PR #490, opened 2019-06-10
Stop and start polling to ensure ballot is printed.

### LINKED ISSUE #484: After markVoterCardUsed() cast page is displayed before ballot can be printed.

beausmith:
Fixes #484

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -183,47 +183,50 @@ export class App extends React.Component<RouteComponentProps, State> {
   }
 
   public startPolling = () => {
-    let lastCardDataString = ''
+    if (checkCardInterval === 0) {
+      let lastCardDataString = ''
 
-    checkCardInterval = window.setInterval(() => {
-      fetch('/card/read')
-        .then(result => result.json())
-        .then(card => {
-          // check whether this has changed to reduce work by a lot
-          const currentCardDataString = JSON.stringify(card)
-          if (currentCardDataString === lastCardDataString) {
-            return
-          }
-          lastCardDataString = currentCardDataString
+      checkCardInterval = window.setInterval(() => {
+        fetch('/card/read')
+          .then(result => result.json())
+          .then(card => {
+            // check whether this has changed to reduce work by a lot
+            const currentCardDataString = JSON.stringify(card)
+            if (currentCardDataString === lastCardDataString) {
+              return
+            }
+            lastCardDataString = currentCardDataString
 
-          const { isVoterCardPresent } = this.state
-          if (isVoterCardPresent && !card.present) {
-            this.resetBallot()
-            return
-          }
+            const { isVoterCardPresent } = this.state
+            if (isVoterCardPresent && !card.present) {
+              this.resetBallot()
+              return
+            }
 
-          if (card.shortValue) {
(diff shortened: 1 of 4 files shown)


benadida:
I'm worried this
- won't fix the issue we're seeing, and
- commits us further down the path of using state on '/', which is feeling a bit weak.

Let's discuss?

beausmith:
Discussed. Was able to reproduce issue. Fixed.

beausmith:
@benadida - Forced-pushed change we verbally discussed. Please review.

src/pages/PrintPage.tsx
```diff
@@ -17,22 +17,22 @@ interface State {
...
-    window.removeEventListener('afterprint', this.resetBallot)
+    window.removeEventListener('afterprint', this.afterPrint)
   }
-  public resetBallot = () => {
+  public afterPrint = () => {
     // setTimeout to prevent a React infinite recursion issue
     window.setTimeout(() => {
       this.context.incrementBallotsPrintedCount()
-      this.context.resetBallot('/cast')
+      this.context.resetBallot('/')
```
benadida: Why do we need to do this, it feels better to go `/cast` for the screen and then not use the state anymore to alter what `/` does?

src/App.tsx
```diff
@@ -392,6 +400,7 @@ export class App extends React.Component<RouteComponentProps, State> {
       this.setStoredState()
       this.props.history.push(path)
     })
+    this.startPolling()
```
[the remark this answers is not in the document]
beausmith: It doesn't need to be as there is no need to wait until the `userState` is updated before starting to poll again. The callback to setState only ensures that the callback happens after the state is updated. The polling will trigger `proccessCardData` which will write over the necessary state. 

Because the callback it unrelated to the polling, lets keep them separate.

src/pages/PrintPage.tsx
beausmith: It's not possible to use the router to direct based on the url because there is no router loaded in this state. This was the workaround which we decided on late night last week. I had simply forgot to remove the route. This routing is completely useless at this point, hence I was removing it.

src/App.tsx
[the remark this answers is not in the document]
benadida: cool, 👍

src/pages/PrintPage.tsx
benadida: Right, I'm wondering if the solution from last week is no longer the way we want to go given this new approach of no longer polling. I'm wondering if we should direct to `/cast` and render the fourth interstitial there.

beausmith: Removing the `/cast` url is simply clean up from the previous changes.

In order to render at a specific URL we would need to add a second router (similar to what is in `Ballot.tsx`) and then manage paths (of which there would only be `/` and `/cast` at this point).

Thus the solution you are suggesting involves more work which is not worth it until we need to navigate multiple pages in this at different URLs in this "logged out" state.

Any change to the strategy should be captured in a new issue, no?

benadida: I can see your point, though I still worry about having `/` do anything other than start the election. I can live with where we are today.

PR #493, opened 2019-06-11
Use pointer events for faster response.

### LINKED ISSUE #334: Handle all presses: long and short
## Problem
- Tap and hold on candidates doesn’t work.
- Short press displays a cursor but doesn't register as a tap.
- If first tap doesn't work, voters tend to push harder and longer to indicate their intention, which registers a long press.

## Solutions
- Capture short tap as normal press. 
- Capture long press as normal press.

## Implementation
- [x] Use `onPointerUp` GlobalEventHandler

beausmith:
Fixes #334

### DIFF
--- a/src/components/Ballot.tsx
+++ b/src/components/Ballot.tsx
@@ -23,10 +23,10 @@ const Ballot = () => {
   const { showHelpPage, showSettingsPage } = election!.bmdConfig!
 
   const onActive = () => {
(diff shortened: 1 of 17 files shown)


beausmith:
Try using `onClick` and you'll get a TypeScript error! :)

benadida:
> Try using `onClick` and you'll get a TypeScript error! :)

Nice.

PR #499, opened 2019-06-12
Restore fully functionaly keyboard accessibility.

beausmith:
Fixes #498

PR #503, opened 2019-06-12
Buttons all the way down

### LINKED ISSUE #500: On contests: change label/checkbox combo to simple button to work with touchevents `onPointerUp`

beausmith:
Fixes #500 

Note: this PR is branched off of: https://github.com/votingworks/bmd/pull/499

### DIFF
--- a/src/components/CandidateContest.tsx
+++ b/src/components/CandidateContest.tsx
@@ -21,6 +21,7 @@ import {
 import BallotContext from '../contexts/ballotContext'
 
 import GLOBALS from '../config/globals'
+import ChoiceButton from './ChoiceButton'
 import Button from './Button'
 import Main from './Main'
 import Modal from './Modal'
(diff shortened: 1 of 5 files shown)


src/components/CandidateContest.tsx
```diff
@@ -297,16 +245,25 @@ class CandidateContest extends React.Component<Props, State> {
   }
 
   public closeAttemptedVoteAlert = () => {
-    this.setState({ attemptedOvervoteCandidate: undefined })
+    // Delay to avoid passing tap to next screen
```
benadida: this is gonna be painful to have everywhere and test for.

src/lib/gamepad.ts
```diff
@@ -47,24 +45,20 @@ function handleArrowDown() {
 function handleArrowLeft() {
   const prevButton = document.getElementById('previous') as HTMLButtonElement
   if (prevButton) {
-    prevButton.dispatchEvent(upPointerEvent)
+    prevButton.dispatchEvent(pointerUpEvent)
```
benadida: now that i look at this... is the right thing to dispatch a const event, rather than minting a new one? I wonder if the latter isn't more "by the book".

src/components/CandidateContest.tsx
beausmith: Yes. Hoping to have a better solution based upon how much testing pain this causes.

src/lib/gamepad.ts
beausmith: Just so that I understand… are you suggesting this…?

```diff
- prevButton.dispatchEvent(pointerUpEvent)
+ prevButton.dispatchEvent(new PointerEvent('pointerup', { bubbles: true }))
```

If so, what difference does it make? If not, what are you suggesting?

benadida: That's what I'm suggesting yes. I'm wondering if there are internal event implementation reasons why they shouldn't be reused. Not certain, just being paranoid. Like what happens if the event is marked "no bubbling" at some later point?

PR #506, opened 2019-06-13
Card not present logic resets Ballot correctly.

### LINKED ISSUE #495: Only show Cast page if the cardData has `bp: 1`

beausmith:
Fixes #495


benadida:
reviewed together, good to go.

PR #529, opened 2019-06-21
Alternative fast click - Does it work?

### LINKED ISSUE #514: Alternative fast-click approach
While reading docs on the `touch-action` css property, I came across a SO post ([Eliminate 300ms delay on click events in mobile Safari](https://stackoverflow.com/a/21062165/101290)) which then led me to:
- https://webkit.org/blog/5610/more-responsive-tapping-on-ios/
- https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away
- https://developers.google.com/web/fundamentals/performance/rail

Which then lead me to want to test this alternate path for implementing fast click which currently uses `onPointerUp`… test if the following two changes will trigger clicks as fast as the current `onPointerUp` solution:
- [x] Update Button styles to include `touch-action: manipulation`:
- [x] Replace Button `onPointerUp` with `onClick`

**Note:** Testing this requires access to an **ēlo** touchscreen in a dev env.

beausmith:
Fixes #514 - Only the last commit on this branch (other commits are from currently open PRs)

@benadida - In making progress on test coverage, I ran into an issue with JSDOM not being aware of pointer events yet. Grrr. So, I'd like to ask if you can test this branch on a device with the touch screen attached to see if these changes still produce the fast touch response that `onPointerUp` does? Read more here: https://github.com/votingworks/bmd/issues/514

### DIFF
--- a/cypress/support/index.js
+++ b/cypress/support/index.js
@@ -13,5 +13,5 @@
 // https://on.cypress.io/configuration
 // ***********************************************************
 
-import 'cypress-testing-library/add-commands'
+import '@testing-library/cypress/add-commands'
 import './commands'
--- a/package.json
+++ b/package.json
@@ -109,11 +109,11 @@
     "wait-for-expect": "^1.2.0"
   },
   "devDependencies": {
+    "@testing-library/cypress": "^4.0.3",
     "@testing-library/react": "^8.0.1",
     "@typescript-eslint/eslint-plugin": "^1.9.0",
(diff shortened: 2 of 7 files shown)


benadida:
this is working just fine on the touchscreen.

beausmith:
@benadida - FYI, I force-pushed to this branch after your 👍 as I had missed the replacement of one `onPointerUp`. I'm confident in this change as the testing branch tests this code.

PR #626, opened 2019-08-15
refactor: clean up types in `isJSON` utils

### LINKED ISSUE #1: break up into separate files by component

eventualbuddha:
This gets rid of `any` and replaces it with [`unknown`][unknown] as appropriate. Using `unknown` is better because it is the _least_ capable type rather than `any` which is the _most_ capable type. This commit changes `isString` and `isObject` to be [type predicates][type predicates] which allows TS to infer types of arguments based on the return value, so `unknown` can be implicitly cast to `string` or `object` when control flow is based on the results of `isString` and `isObject`. Note that `isString` is not really necessary since `typeof value === 'string'` does the same thing, but I left it to minimize changes to this file.

`passObject` now has a default value of `false` just to simplify its type from `boolean | undefined` to just `boolean` while keeping the API the same.

Using `.map((s: any) => isJSON(s)).reduce((prev: any, curr: any) => !!curr)` is not good for a few reasons:
- all those `any` type annotations are unnecessary
- using `Array#reduce` where `prev` is ignored is a hint that there's a more appropriate method to use, in this case `Array#every`
- using `!!` on array entries is redundant when they are guaranteed to be `boolean` values because they come from `isJSON`

[unknown]: https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type
[type predicates]: http://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

### DIFF
--- a/src/utils/isJSON.ts
+++ b/src/utils/isJSON.ts
@@ -1,50 +1,46 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
 // Ported to TypeScript from:
 // https://github.com/joaquimserafim/isJSON
 
-function isString(x: string) {
+function isString(x: unknown): x is string {
   return Object.prototype.toString.call(x) === '[object String]'
 }
 
-function isObject(obj: {}) {
+function isObject(obj: unknown): obj is object {
   return Object.prototype.toString.call(obj) === '[object Object]'
 }
 
-function isJSON(str: any, passObject?: boolean): boolean {
-  if (passObject && isObject(str)) {
+function isJSON(value: unknown, passObject: boolean = false): boolean {
+  if (passObject && isObject(value)) {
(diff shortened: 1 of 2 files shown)


src/utils/isJSON.ts
```diff
@@ -1,50 +1,46 @@
...
     return true
   }
 
-  if (!isString(str)) {
+  if (!isString(value)) {
     return false
   }
 
-  str = str.replace(/\s/g, '').replace(/\n|\r/, '')
+  const str = value.replace(/\s/g, '').replace(/\n|\r/, '')
```
eventualbuddha: One thing I didn't note in the commit message is that I changed this to avoid reassigning a parameter.

PR #629, opened 2019-08-15
Long live the tests

beausmith:
'tis possible cypress tests may fail.


src/AppCard.test.tsx
```diff
@@ -1,157 +1,139 @@
 import React from 'react'
-import ReactDOM from 'react-dom'
+// import ReactDOM from 'react-dom'
```
eventualbuddha: Remove commented out code?

PR #650, opened 2019-08-20
Upgrade to latest CRA

beausmith:
This should take care of the security issue in #576, and resolve #648 too.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,10 +42,10 @@ module.exports = {
   plugins: [
     '@typescript-eslint',
     'jest',
-    'no-null',
     'react',
     'cypress',
     'jsx-a11y',
+    'no-null',
   ],
   settings: {
     react: {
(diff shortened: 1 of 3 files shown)


eventualbuddha:
I think I know how to fix the issues flagged by CI on this branch, but unfortunately it's not super easy. Since `react-scripts` has its own dependency on `eslint`, the version they depend on can get out of sync with the one we're using. So we need to update to `>= 6`, with v6.2.1 being the latest.

That reveals other issues. It seems eslint v6 is stricter about the config file contents, which shows that the rule `@typescript-eslint/no-null-keyword` that we enable does not exist. Additionally, `eslint:recommended` now includes more rules such as `no-shadow` and `no-nested-ternary`.

I've worked through most of these, but I went a little overboard trying to debug things. I'll unwind my changes to just enough to get this branch to pass and submit a PR to this PR.

PR #651, opened 2019-08-20
Fix latest CRA

eventualbuddha:
Includes fixes for #650 to make it pass.


eventualbuddha:
Okay, I seem to continue hitting issues even if I add `EXTEND_ESLINT` to everything. Any suggestions to fix this, @beausmith?

eventualbuddha:
Too old and broken to fix, will try again later.

PR #694, opened 2019-09-05
App mode (to support VxMark, VxPrint, VxPrint+Mark)

beausmith:
Related #695


src/config/types.ts
```diff
@@ -4,8 +4,12 @@ export interface Dictionary<T> {
   [key: string]: T | undefined
 }
 
+// App
+export type AppMode = 'mark' | 'print' | 'mark+print'
```
eventualbuddha: I might have done this as a sort of configuration object. That way you don't have to enumerate the mode names everywhere and encompassing a feature is easier, i.e. you don't have to say `appMode === 'print' || appMode === 'mark+print'`, you can just say `appMode.isVxPrint`. This is more useful when you have more than three configurations, but I think it still has merit here. I tried it locally and found I couldn't entirely get rid of the mode names due to serialization. Here's the diff:

```diff
diff --git a/src/AppRoot.tsx b/src/AppRoot.tsx
index 0d2f369..af9ac42 100644
--- a/src/AppRoot.tsx
+++ b/src/AppRoot.tsx
@@ -20,7 +20,6 @@ import {
   Contests,
   Election,
   ElectionDefaults,
-  InputEvent,
   MachineIdAPI,
   OptionalElection,
   OptionalVote,
@@ -28,6 +27,8 @@ import {
   UserSettings,
   VoterCardData,
   VotesDict,
+  VxMarkOnly,
+  getAppMode,
 } from './config/types'
 
 import utcTimestamp from './utils/utcTimestamp'
@@ -105,7 +106,7 @@ const initialUserState: UserState = {
 const initialState: State = {
   ...initialUserState,
   ...initialCardPresentState,
-  appMode: 'mark',
+  appMode: VxMarkOnly,
   ballotsPrintedCount: 0,
   election: undefined,
   isLiveMode: false,
@@ -447,9 +448,21 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
     )
   }
 
-  public getStoredState = () => {
-    const storedState = window.localStorage.getItem(stateStorageKey)
-    return storedState ? JSON.parse(storedState) : {}
+  public getStoredState = (): Partial<State> => {
+    const storedStateJSON = window.localStorage.getItem(stateStorageKey)
+    const storedState: Partial<State> = storedStateJSON
+      ? JSON.parse(storedStateJSON)
+      : {}
+
+    if (storedState.appMode) {
+      try {
+        storedState.appMode = getAppMode(storedState.appMode.name)
+      } catch {
+        delete storedState.appMode
+      }
+    }
+
+    return storedState
   }
 
   public updateVote = (contestId: string, vote: OptionalVote) => {
@@ -520,9 +533,7 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
     document.documentElement.style.fontSize = `${GLOBALS.FONT_SIZES[textSize]}px`
   }
 
-  public setAppMode = (event: InputEvent) => {
-    const currentTarget = event.currentTarget as HTMLInputElement
-    const appMode = currentTarget.dataset.appMode as AppMode
+  public setAppMode = (appMode: AppMode) => {
     this.setState({ appMode }, this.setStoredState)
   }
 
diff --git a/src/config/types.ts b/src/config/types.ts
index d6bd5a2..2286569 100644
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -5,7 +5,35 @@ export interface Dictionary<T> {
 }
 
 // App
-export type AppMode = 'mark' | 'print' | 'mark+print'
+export interface AppMode {
+  readonly name: string
+  readonly isVxPrint?: boolean
+  readonly isVxMark?: boolean
+}
+
+export const VxPrintOnly: AppMode = { name: 'print', isVxPrint: true }
+export const VxMarkOnly: AppMode = { name: 'mark', isVxMark: true }
+export const VxMarkPlusVxPrint: AppMode = {
+  name: 'mark+print',
+  isVxPrint: true,
+  isVxMark: true,
+}
+
+export function getAppMode(name: string): AppMode {
+  switch (name) {
+    case VxPrintOnly.name:
+      return VxPrintOnly
+
+    case VxMarkOnly.name:
+      return VxMarkOnly
+
+    case VxMarkPlusVxPrint.name:
+      return VxMarkPlusVxPrint
+
+    default:
+      throw new Error(`unknown app mode: ${name}`)
+  }
+}
 
 // Events
 export type InputEvent = React.FormEvent<EventTarget>
diff --git a/src/contexts/ballotContext.ts b/src/contexts/ballotContext.ts
index 2041a4b..c738c64 100644
--- a/src/contexts/ballotContext.ts
+++ b/src/contexts/ballotContext.ts
@@ -1,11 +1,15 @@
 import { createContext } from 'react'
 import * as GLOBALS from '../config/globals'
 
-import { BallotContextInterface, TextSizeSetting } from '../config/types'
+import {
+  BallotContextInterface,
+  TextSizeSetting,
+  VxMarkOnly,
+} from '../config/types'
 
 const ballot: BallotContextInterface = {
   activateBallot: () => undefined,
-  appMode: 'mark',
+  appMode: VxMarkOnly,
   ballotStyleId: '',
   contests: [],
   election: undefined,
diff --git a/src/pages/ClerkScreen.test.tsx b/src/pages/ClerkScreen.test.tsx
index 8b5d4ec..a40f893 100644
--- a/src/pages/ClerkScreen.test.tsx
+++ b/src/pages/ClerkScreen.test.tsx
@@ -9,6 +9,7 @@ import { election } from '../../test/helpers/election'
 import { noCard, adminCard, advanceTimers } from '../../test/helpers/smartcards'
 
 import ClerkScreen from './ClerkScreen'
+import { VxMarkOnly } from '../config/types'
 
 jest.useFakeTimers()
 
@@ -18,7 +19,7 @@ fetchMock.get('/card/read', () => JSON.stringify(currentCard))
 it('renders ClerkScreen', async () => {
   const { getByText } = render(
     <ClerkScreen
-      appMode="mark"
+      appMode={VxMarkOnly}
       ballotsPrintedCount={0}
       election={election}
       fetchElection={jest.fn()}
diff --git a/src/pages/ClerkScreen.tsx b/src/pages/ClerkScreen.tsx
index af6da1a..74e053a 100644
--- a/src/pages/ClerkScreen.tsx
+++ b/src/pages/ClerkScreen.tsx
@@ -2,9 +2,11 @@ import React, { useState } from 'react'
 
 import {
   AppMode,
-  InputEventFunction,
   OptionalElection,
   VoidFunction,
+  VxMarkOnly,
+  VxPrintOnly,
+  VxMarkPlusVxPrint,
 } from '../config/types'
 
 import TestBallotDeckScreen from './TestBallotDeckScreen'
@@ -23,7 +25,7 @@ interface Props {
   isLiveMode: boolean
   fetchElection: VoidFunction
   isFetchingElection: boolean
-  setAppMode: InputEventFunction
+  setAppMode: (appMode: AppMode) => void
   toggleLiveMode: VoidFunction
   unconfigure: VoidFunction
 }
@@ -61,10 +63,7 @@ const ClerkScreen = ({
       />
     )
   }
-  const isVxMark = appMode === 'mark'
-  // const isVxPrint = appMode === 'print'
-  // const isVxMarkAndPrint = appMode === 'mark+print'
-  const isTestDecksAvailable = isLiveMode || (!isLiveMode && isVxMark)
+  const isTestDecksAvailable = isLiveMode || (!isLiveMode && appMode.isVxMark)
   return (
     <React.Fragment>
       <Main>
@@ -77,26 +76,23 @@ const ClerkScreen = ({
                 <p>This device can operate as multiple apps.</p>
                 <SegmentedButton>
                   <Button
-                    onPress={setAppMode}
-                    primary={appMode === 'mark'}
-                    disabled={appMode === 'mark'}
-                    data-app-mode="mark"
+                    onPress={() => setAppMode(VxMarkOnly)}
+                    primary={appMode === VxMarkOnly}
+                    disabled={appMode === VxMarkOnly}
                   >
                     VxMark
                   </Button>
                   <Button
-                    onPress={setAppMode}
-                    primary={appMode === 'print'}
-                    disabled={appMode === 'print'}
-                    data-app-mode="print"
+                    onPress={() => setAppMode(VxPrintOnly)}
+                    primary={appMode === VxPrintOnly}
+                    disabled={appMode === VxPrintOnly}
                   >
                     VxPrint
                   </Button>
                   <Button
-                    onPress={setAppMode}
-                    primary={appMode === 'mark+print'}
-                    disabled={appMode === 'mark+print'}
-                    data-app-mode="mark+print"
+                    onPress={() => setAppMode(VxMarkPlusVxPrint)}
+                    primary={appMode === VxMarkPlusVxPrint}
+                    disabled={appMode === VxMarkPlusVxPrint}
                   >
                     VxMark+Print
                   </Button>
@@ -133,18 +129,18 @@ const ClerkScreen = ({
                       (Available in testing mode)
                     </Text>
                   )}
-                  {!isLiveMode && isVxMark && (
+                  {!isLiveMode && !appMode.isVxPrint && (
                     <Text as="small" muted>
                       (Available with VxPrint or VxMark+Print)
                     </Text>
                   )}
                 </p>
-                <Text as="h1" muted={isVxMark}>
+                <Text as="h1" muted={!appMode.isVxPrint}>
                   Stats
                 </Text>
-                <Text muted={isVxMark}>
+                <Text muted={!appMode.isVxPrint}>
                   Printed Ballots: <strong>{ballotsPrintedCount}</strong>{' '}
-                  {isVxMark && (
+                  {!appMode.isVxPrint && (
                     <small>(Available with VxPrint or VxMark+Print)</small>
                   )}
                 </Text>
diff --git a/src/pages/PollWorkerScreen.tsx b/src/pages/PollWorkerScreen.tsx
index 63a4c7d..2ba2731 100644
--- a/src/pages/PollWorkerScreen.tsx
+++ b/src/pages/PollWorkerScreen.tsx
@@ -93,13 +93,11 @@ const PollWorkerScreen = ({
   const showModal = () => setIsModalOpen(true)
   const hideModal = () => setIsModalOpen(false)
 
-  const isVxPrint = appMode === 'print'
-  const isVxMarkAndPrint = appMode === 'mark+print'
   const currentDateTime = new Date().toLocaleString()
   const numReports = 3
 
   const togglePollsOpen = () => {
-    if (isVxPrint || isVxMarkAndPrint) {
+    if (appMode.isVxPrint) {
       showModal()
     } else {
       appTogglePollsOpen()
diff --git a/src/pages/ReviewPage.tsx b/src/pages/ReviewPage.tsx
index cd31974..264a8f4 100644
--- a/src/pages/ReviewPage.tsx
+++ b/src/pages/ReviewPage.tsx
@@ -360,7 +360,7 @@ class ReviewPage extends React.Component<RouteComponentProps, State> {
           </ContentFooter>
         </Main>
         <ButtonBar>
-          <LinkButton to={appMode === 'mark' ? '/remove' : '/print'} id="next">
+          <LinkButton to={!appMode.isVxPrint ? '/remove' : '/print'} id="next">
             Next
           </LinkButton>
           <div />
diff --git a/src/utils/utcTimestamp.ts b/src/utils/utcTimestamp.ts
index e818f54..03dc41c 100644
--- a/src/utils/utcTimestamp.ts
+++ b/src/utils/utcTimestamp.ts
@@ -1,3 +1,5 @@
-// returns current UTC unix timestamp (epoch) in seconds
+/**
+ * Returns current UTC unix timestamp (epoch) in seconds
+ */
 const utcTimestamp = (): number => Math.round(Date.now() / 1000)
 export default utcTimestamp
diff --git a/test/testUtils.tsx b/test/testUtils.tsx
index 0481290..48b69a8 100644
--- a/test/testUtils.tsx
+++ b/test/testUtils.tsx
@@ -11,10 +11,10 @@ import * as GLOBALS from '../src/config/globals'
 import electionSampleNoSeal from '../src/data/electionSampleNoSeal.json'
 
 import {
-  AppMode,
   Contests,
   Election,
   TextSizeSetting,
+  VxMarkOnly,
 } from '../src/config/types'
 
 import { mergeWithDefaults } from '../src/AppRoot'
@@ -25,7 +25,7 @@ export function render(
   {
     route = '/',
     activateBallot = jest.fn(),
-    appMode = 'mark' as AppMode,
+    appMode = VxMarkOnly,
     ballotStyleId = '',
     contests = electionSampleNoSeal.contests as Contests,
     markVoterCardUsed = jest.fn(),
```

src/pages/ActivationScreen.tsx
```diff
@@ -16,36 +16,47 @@ const InsertCardImage = styled.img`
...
-              <p>This card is no longer active.</p>
-              <p>Please return card to poll worker.</p>
+              <p>Please return this card to a poll worker.</p>
             </React.Fragment>
-          ) : (
+          )}
+          {showExpired && (
             <React.Fragment>
+              <h1>Expired Card</h1>
+              <p>Please return this card to a poll worker for assistance.</p>
```
eventualbuddha: This copy differs from the "Inactive Card" case. Should it?

src/utils/utcTimestamp.ts
```diff
@@ -0,0 +1,3 @@
+// returns current UTC unix timestamp (epoch) in seconds
```
eventualbuddha: If you use JSDoc then you get this comment in your editor intellisense:

```suggestion
/**
 * Returns current UTC unix timestamp (epoch) in seconds
 */
```

image

src/pages/ActivationScreen.tsx
beausmith: Good question. Yeah, I think something can be simplified here. 

There are two different states which perhaps we do not need to differentiate between for voters/poll workers. The resolution is the same for the Poll Worker, they simply re-encode the card. Though we could update the VxEncode UI to give the Poll Worker more information about the state of the card.

### Inactive card
- A voter card which has printed a ballot.
- If found and inserted into a BMD, the UI should display message that this card should be returned to the poll worker to reenter circulation.
- I'm trying to avoid using the copy "Used card" such that the card is tossed in the bin.

### Expired card
- A voter card which has not been used to print a ballot, but for which the voter too more than 10 minutes to insert into a BMD after being encoded.
- A voter should be able to re-activate this card with the help of a Poll Worker.

Any further thoughts based upon this info?

src/utils/utcTimestamp.ts
beausmith: Awesome, will update.

src/config/types.ts
beausmith: Thanks for your thoughts on this. I went back and forth on a couple ways to implement this. Happy to use this solution. Do you have this stashed somewhere which would make it easier to apply these changes?

eventualbuddha: Yep, here's the commit: https://github.com/votingworks/bmd/commit/def1d6071dbd2646db38707eb79dcb46e4bbb65f

src/pages/ActivationScreen.tsx
eventualbuddha: Thanks for the context. I think this is fine as a first pass. We can always iterate.

src/pages/ClerkScreen.tsx
```diff
@@ -61,10 +64,7 @@ const ClerkScreen = ({
       />
     )
   }
-  const isVxMark = appMode === 'mark'
-  // const isVxPrint = appMode === 'print'
-  // const isVxMarkAndPrint = appMode === 'mark+print'
-  const isTestDecksAvailable = isLiveMode || (!isLiveMode && isVxMark)
+  const isTestDecksAvailable = isLiveMode || (!isLiveMode && appMode.isVxMark)
```
eventualbuddha: One thing I found that I incorrectly updated in a few places was changing `appMode === 'mark'` to `appMode.isVxMark` like was done here. I noticed that you were using `appMode === 'mark'` to mean "does not support printing", which means it should be `!appMode.isVxPrint`. Which should be used here?

beausmith: Ok. I reviewed that swap, which is 👍 and I'll re-review the rest to ensure they're correct.

beausmith: Reviewed and amended that one change.

package.json
```diff
@@ -93,7 +93,6 @@
     "history": "^4.9.0",
     "http-proxy-middleware": "^0.20.0",
     "lodash.camelcase": "^4.3.0",
-    "mousetrap": "^1.6.3",
```
eventualbuddha: We can also remove `@types/mousetrap`, right?

src/AppPrintOnly.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+  // Remove card
+  currentCard = noCard
+  advanceTimers()
+  await wait(() => getByText('Insert Card'))
+
+  // ---------------
+
+  // Insert Expired Voter Card
+  // Insert Used Voter Card
+  // Insert Voter Card with No Votes
```
eventualbuddha: Placeholder?

src/AppRoot.tsx
```diff
@@ -221,33 +256,47 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
     checkCardInterval = 0 // To indicate setInterval is not running.
   }
 
-  public markVoterCardUsed = async (
-    { ballotPrinted } = { ballotPrinted: true }
-  ) => {
+  public togglePauseProcessingUntilNoCardPresent = (b?: boolean) => {
```
eventualbuddha: What is `b` here? It seems like passing `true` to this method will force `pauseProcessingUntilNoCardPresent` to be `true`, but passing `false` (or nothing) will toggle the value. This feels a bit confusing. Should these be separate methods? One that toggles and one that just sets it?

src/AppRoot.tsx
```diff
@@ -312,36 +375,48 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
...
-    Mousetrap.unbind(removeElectionShortcuts)
     document.removeEventListener('keydown', handleGamepadKeyboardEvent)
     this.stopPolling()
   }
 
+  public setMachineId = async () => {
+    const { signal } = this.machineIdAbortController
+    try {
+      const response = await fetch('/machine-id', { signal })
+      const { machineId }: MachineIdAPI = await response.json()
```
eventualbuddha: There's a relatively new utility function I added that can shorten this.

```suggestion
      const { machineId } = await fetchJSON<MachineIdAPI>('/machine-id', { signal })
```

src/AppRoot.tsx
```diff
@@ -312,36 +375,48 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
...
+      machineId && this.setState({ machineId })
+    } catch (error) {
+      // TODO: what should happen if `machineId` is not returned?
+    }
+  }
+
+  public readCard = async (): Promise<CardAPI> => {
```
eventualbuddha: This method could be written like this:

```ts
  public readCard = async (): Promise<CardAPI> => {
    return await fetchJSON<CardAPI>('/card/read')
  }
```

src/pages/UsedCardScreen.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import Prose from '../components/Prose'
+import Main, { MainChild } from '../components/Main'
+
+const ExpiredCardScreen = () => (
```
eventualbuddha: copy-pasta

```suggestion
const UsedCardScreen = () => (
```

test/helpers/smartcards.ts
```diff
@@ -1,11 +1,16 @@
...
+import utcTimestamp from '../../src/utils/utcTimestamp'
 
 const election = electionSample as Election
+const contest0 = electionSample.contests[0] as CandidateContest
+const contest1 = electionSample.contests[1] as CandidateContest
+const contest0candidate0 = contest0.candidates[0]
+const contest1candidate0 = contest1.candidates[0]
 
-export const noCard: CardAPI = {
+export const noCard: Partial<CardAPI> = {
```
eventualbuddha: This feels like the types aren't doing enough to help you out. Looking at `CardAPI`, it appears that we want to encode the fact that if `present` is `true` then `shortValue` and maybe also `longValueExists` will be set. We can do that like so:

```diff
diff --git a/src/AppRoot.tsx b/src/AppRoot.tsx
index 5bfb795..5330e34 100644
--- a/src/AppRoot.tsx
+++ b/src/AppRoot.tsx
@@ -35,6 +35,7 @@ import {
   VxMarkPlusVxPrint,
   VxPrintOnly,
   getAppMode,
+  PresentCardAPI,
 } from './config/types'
 
 import utcTimestamp from './utils/utcTimestamp'
@@ -172,7 +173,7 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
   public isVoterCardExpired = (createdAt: number): boolean =>
     utcTimestamp() >= createdAt + GLOBALS.CARD_EXPIRATION_SECONDS
 
-  public processCard = ({ longValueExists, shortValue }: CardAPI) => {
+  public processCard = ({ longValueExists, shortValue }: PresentCardAPI) => {
     const cardData: CardData = JSON.parse(shortValue)
     switch (cardData.t) {
       case 'voter': {
@@ -283,7 +284,7 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
     await this.writeCard(usedVoterCardData)
 
     const updatedCard = await this.readCard()
-    const updatedShortValue: VoterCardData = updatedCard.shortValue
+    const updatedShortValue: VoterCardData = updatedCard.present
       ? JSON.parse(updatedCard.shortValue)
       : {}
 
diff --git a/src/config/types.ts b/src/config/types.ts
index 67065a4..00bf9bf 100644
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -184,11 +184,15 @@ export interface ClerkCardData extends CardData {
   readonly t: 'clerk'
   readonly h: string
 }
-export interface CardAPI {
-  present: boolean
+export type AbsentCardAPI = {
+  present: false
+}
+export type PresentCardAPI = {
+  present: true
   shortValue: string
   longValueExists?: boolean
 }
+export type CardAPI = AbsentCardAPI | PresentCardAPI
 
 // Machine ID API
 export interface MachineIdAPI {
diff --git a/test/helpers/smartcards.ts b/test/helpers/smartcards.ts
index 2d8658d..33b1675 100644
--- a/test/helpers/smartcards.ts
+++ b/test/helpers/smartcards.ts
@@ -10,7 +10,7 @@ const contest1 = electionSample.contests[1] as CandidateContest
 const contest0candidate0 = contest0.candidates[0]
 const contest1candidate0 = contest1.candidates[0]
 
-export const noCard: Partial<CardAPI> = {
+export const noCard: CardAPI = {
   present: false,
 }
 ```

File: package.json (hunk 5 above)
Comment (beausmith) [reply package.json]: yes, will do.

File: src/AppPrintOnly.test.tsx (hunk 6 above)
Comment (beausmith) [reply src/AppPrintOnly.test.tsx]: This is addressed in new commits since your review.

File: src/AppRoot.tsx (hunk 7 above)
Comment (beausmith) [reply src/AppRoot.tsx]: Ah, I see my failure in logic. I could "toggle" only on `undefined`, but since I'm only calling method with a value, I'll remove ability to toggle.

File: src/AppRoot.tsx (hunk 8 above)
Comment (beausmith) [reply src/AppRoot.tsx]: Thx. Will implement.

File: src/pages/UsedCardScreen.tsx (hunk 10 above)
Comment (beausmith) [reply src/pages/UsedCardScreen.tsx]: TSLint had a rule to ensure that default export was named same as filename. I guess I got lazy after relying on it. 😛  Fixing.

File: test/helpers/smartcards.ts (hunk 11 above)
Comment (beausmith) [reply test/helpers/smartcards.ts]: Love it! Makes sense. Thanks for this tip.

**Comment by beausmith** [comment]
@eventualbuddha - Rebased on master. Re-requesting review on most recent 7 commits.

**Comment by beausmith** [comment]
@eventualbuddha - I'm squeezing two more related commits into this PR since it still has not been re-reviewed. 😄

File: src/AppRoot.tsx [hunk 12]
```diff
@@ -263,39 +258,60 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
...
     }
+    await this.writeCard(usedVoterCardData)
+
+    const updatedCard = await this.readCard()
+    const updatedShortValue: VoterCardData =
+      updatedCard.present && JSON.parse(updatedCard.shortValue)
 
     this.startPolling()
 
+    /* istanbul ignore next - When the card read doesn't match the card write. Currently not possible to test this without separating the write and read into separate methods and updating printing logic. This is an edige case. */
```
eventualbuddha: `s/edige/edge/`

src/pages/PrintOnlyScreen.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }, [cardVotes, printBallot])
+
+  useEffect(() => {
+    if (!isVoterCardPresent) {
+      updateIsPrinted(false)
+    }
+  }, [isVoterCardPresent])
+
+  useEffect(() => {
+    return () => clearTimeout(printerTimer.current)
```
eventualbuddha: Maybe this can't be avoided in this situation, but one of the selling points of hooks & `useEffect` was that you could co-locate the "register & unregister"-type actions, or in this case `setTimeout` & `clearTimeout`. Is that not possible or not desirable here?

eventualbuddha:
Still LGTM.

PR #699, opened 2019-09-06
don't double handle the click of an Enter key

benadida:
'Enter' key on accessible controller wasn't working because we were artificially handling the click a second time. Let's not do that, 'Enter' does it all by itself


benadida:
> Approved, but wondering does this effect use of the "enter" key on keyboard?

Yeah, in fact the Enter key on the keyboard does *not* work on master today when it's a candidate. It does work with this PR :) The accessible controller's middle button is *exactly* the same as pressing Enter on a keyboard, same USB signal.

PR #700, opened 2019-09-06
better touch via pointer events

benadida:
The more I dig into touch events, the more I think the right thing is to use pointer events instead of click events. Here are the arguments:

* Pointer events are specifically designed to handle both mouse and touch with one type of event: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events . With our current approach -- the click event simulation -- there's no clear spec on what touches qualify as clicks. The difference in behavior we've seen between Chrome on Mac and Linux doesn't appear to be out of spec, it's just a different implementation choice. It's not an OS thing either, since Firefox behaves in our preferred way on Linux, unlike Chrome.

* With click simulation, even when it's the kind we like (e.g. Chrome on Mac), the click isn't issued until the finger is lifted, on `PointerUp`. That doesn't feel nearly as responsive as taking action on `PointerDown`. Try the `PointerDown` behavior with a touchscreen... it feels *wonderfully* responsive. I wonder if that's one reason why we've seen some testers hold their finger down longer, because they're not seeing an immediate reaction on touch.

* I'm not seeing any unexpected behavior with this approach. You can hold your finger as long as you want, a second `PointerDown` event isn't issued. You have to lift your finger and retouch... which would be a second click anyways in our current model. No timeouts, no special handling. It just works right, as best as I can tell.

* looks like we can change our tests to use `fireEvent.pointerDown` and all works well. See the one example in this PR.

### DIFF
--- a/src/components/Button.test.tsx
+++ b/src/components/Button.test.tsx
@@ -62,17 +62,24 @@ it('works properly with clicks and touches', () => {
   fireEvent.pointerDown(button)
   expect(onPress).toHaveBeenCalledTimes(2)
 
-  // this is the behavior when a proper touch becomes a click
+  // when a tap is not smudged and doesn't last too long
+  // we get pointerDown, pointerUp, and finally a click event.
+  // in this case, we only want onPress to fire once.
   fireEvent.pointerDown(button)
   fireEvent.pointerUp(button)
   fireEvent.click(button)
   expect(onPress).toHaveBeenCalledTimes(3)
 
-  // this is the behavior on a bad touch, followed by a separate click
-  // so this ends up calling the event handler twice
+  // when a tap is smudged or lasts too long,
+  // we get pointerDown, pointerCancel, and no click event.
+  // we still want onPress to fire exactly once.
   fireEvent.pointerDown(button)
   fireEvent.pointerCancel(button)
   expect(onPress).toHaveBeenCalledTimes(4)
+
+  // on use of accessible controller / keyboard
+  // we get just a click event.
+  // this should trigger onPress exactly once.
   fireEvent.click(button)
   expect(onPress).toHaveBeenCalledTimes(5)
 })
--- a/src/components/Button.test.tsx
+++ b/src/components/Button.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 
 import Button, { DecoyButton } from './Button'
 
@@ -49,4 +49,29 @@ it('renders small Button', () => {
 it('renders DecoyButton', () => {
   const { container } = render(<DecoyButton>DecoyButton</DecoyButton>)
   expect(container.firstChild).toMatchSnapshot()
+})
+
+it('works properly with clicks and touches', () => {
+  const onPress = jest.fn()
+  const { getByText } = render(<Button onPress={onPress}>Test Button</Button>)
+  const button = getByText('Test Button')
+
+  fireEvent.click(button)
+  expect(onPress).toHaveBeenCalledTimes(1)
+
+  fireEvent.pointerDown(button)
+  expect(onPress).toHaveBeenCalledTimes(2)
+
+  // this is the behavior when a proper touch becomes a click
+  fireEvent.pointerDown(button)
+  fireEvent.pointerUp(button)
+  fireEvent.click(button)
+  expect(onPress).toHaveBeenCalledTimes(3)
+
+  // this is the behavior on a bad touch, followed by a separate click
+  // so this ends up calling the event handler twice
+  fireEvent.pointerDown(button)
+  fireEvent.pointerCancel(button)
+  fireEvent.click(button)
+  expect(onPress).toHaveBeenCalledTimes(5)
 })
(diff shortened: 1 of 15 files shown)


beausmith:
Awesome! I'm convinced, especially if this is true:

> The difference in behavior we've seen between Chrome on Mac and Linux doesn't appear to be out of spec, it's just a different implementation choice.

eventualbuddha:
Seems good to me!

benadida:
@eventualbuddha @beausmith ok, I changed my approach again.

Why? Because moving to just pointer events messed with the entire approach to accessibility, not just `next` and `prev`, since we can't just `click` an element and also space-bar and enter no longer work, since those simulate click events, which are no longer handled. Boo 😢 😱 

So instead I am trying the following approach:
- `onpointerdown`: handle the press
- `onpointerup`: set an internal state flag to not process the next click event. This is the right time to set this flag, because `onpointerup` only triggers after `onpointerdown`, and it's a tell-tale sign that `onclick` is about to fire.
- `onclick`: if the flag is *not* set, handle the press. If it is set, unset it for the next time around.

As best as I can tell, this is working marvelously with very little code changes: whichever of `pointerdown` or `click` fires first is processed, and `click` still works for accessibility purposes and for testing.

So there's an argument to be made that the tests should intermix `pointerDown` and `click` now, which I'll do if you think this is a good direction.

benadida:
OK, @beausmith @eventualbuddha I've tried everything else, and the only way I can find that seems "nice" is the current PR, where buttons handle both `click` and `pointerdown`, but on `pointerup` set a flag to prevent `click` from being handled and duplicating the action of `pointerdown`.

There's a remaining behavior where, after clicking through, if the finger is still on the screen for 0.5s or more, the button on the next screen gets the dotted-line outline. It doesn't get activated, so this still feels like progress, but I'd still like to eventually prevent that from happening. I'll work on it. In the meantime, I think this PR is good to go.

src/AppEndToEnd.test.tsx
```diff
@@ -253,27 +259,30 @@ it('basic end-to-end flow', async () => {
   getByText(/Vote for 4/i)
 
   // Select first candidate
```
eventualbuddha: This comment seems incorrect. It's selecting the first _and_ second candidates, right?

src/AppEndToEnd.test.tsx
```diff
@@ -221,21 +221,27 @@ it('basic end-to-end flow', async () => {
...
     else if (title === measure102Contest.title) {
-      fireEvent.click(getByText('Yes'))
+      fireEvent.pointerDown(getByText('Yes'))
     }
 
-    fireEvent.click(getByText('Next'))
+    const button = getByText('Next')
+    fireEvent.pointerDown(button)
+    fireEvent.pointerUp(button)
+    fireEvent.click(button)
```
eventualbuddha: Are these necessary here, or are you just trying to make sure both work? Should that be its own test so as not to be confusing here?

src/AppEndToEnd.test.tsx
beausmith: Previous copy/pasta by me!

src/AppEndToEnd.test.tsx
beausmith: Perhaps an exhaustive standalone `Button` test which confirms only the expectations of the button actions should exist? 

All the other tests can then remain the same (using `onClick`).

benadida:
@eventualbuddha good callout. I think it's the right tradeoff since we've got the scrolling buttons. I wonder if we might want to turn off swipe-to-scroll in general?

src/AppEndToEnd.test.tsx
benadida: If we're going to explicitly support both for the whole flow, we probably want to make sure we test e2e flows with both approaches. I can imagine we change the component, for example, and then lose the behavior we need in a way that's invisible for tests.

Maybe have the same e2e test run twice, once with pointerDown, once with click.

eventualbuddha:
> I wonder if we might want to turn off swipe-to-scroll in general?

Yes, absolutely. Otherwise anyone who is used to scrolling on a touch device (i.e. anyone with a iOS / Android smartphone or tablet) will likely be frustrated. I still don't think this is necessarily the right trade-off, but if we're going to do it I don't think we can let the _same touch_ be used for both selection and scrolling.

src/AppEndToEnd.test.tsx
eventualbuddha: Running twice isn't necessarily a bad idea.

src/components/Button.test.tsx
```diff
@@ -50,3 +50,29 @@ it('renders DecoyButton', () => {
...
+  fireEvent.click(button)
+  expect(onPress).toHaveBeenCalledTimes(3)
+
+  // this is the behavior on a bad touch, followed by a separate click
+  // so this ends up calling the event handler twice
+  fireEvent.pointerDown(button)
+  fireEvent.pointerCancel(button)
+  expect(onPress).toHaveBeenCalledTimes(4)
+  fireEvent.click(button)
+  expect(onPress).toHaveBeenCalledTimes(5)
```
beausmith: Can I get a little more clarity with the desired behavior here?

From the comments on the last two test blocks… I understand that the "proper touch" is what we expect/want to happen (one call to `onPress`) and the "bad touch" is what we don't want to happen (two calls to `onPress`)?

Can you update the comments in this test case to be explicit about what the desired behavior is?

If this "bad touch" order of events is something we wish to fix, can you create an issue and link to it here in the comment?

beausmith:
Related PRs:
- https://github.com/votingworks/bmd/pull/718
- https://github.com/votingworks/bmd/pull/719

PR #707, opened 2019-09-12
docs: describe the ballot encoding process

eventualbuddha:
This document describes the encoding format for ballots to be used by QR codes and smart cards. At the moment it focuses only on encoding the votes themselves and still needs the following:

- [ ] VotingWorks [file signature](https://en.wikipedia.org/wiki/List_of_file_signatures) (i.e. `VX`)
- [ ] encoding version number
- [ ] ballot style id
- [ ] precinct id
- [ ] ballot id

And maybe more?

([Rendered](https://github.com/votingworks/bmd/blob/262dedf07741affe41fbb9915484307162da1671/docs/encoding.md))


eventualbuddha:
This has been incorporated into votingworks/ballot-encoder#1.

Issue #485, opened 2019-06-10
Display "Ballot ID" on printed ballot to make it easy to find a specific paper ballot when necessary.


beausmith:

benadida:
I will follow up to make the call here based on RLA vs. various state rules.

benadida:
for now, let's not print it. Not useful. Reopen later if we need to.

beausmith:
Ballot ID is needed to find a particular ballot in the case that one is not scanned by VxScan.

benadida:
let's think about how we can make sure this doesn't imply that it identifies the voter.

Issue #424, opened 2019-05-31
Create Unique IDs for ballots


beausmith:
The `ballotID` is currently generated in the `PrintedBallot` component.

Fix in:
- `src/pages/PrintPage.tsx`
- `src/pages/TestBallotDeckScreen.tsx`

benadida:
this is done.

PR #715, opened 2019-09-17
feat: add ballot id back to printed ballot.

beausmith:
Fixes: #485 
Fixes: #424

### DIFF
--- a/src/components/PrintedBallot.tsx
+++ b/src/components/PrintedBallot.tsx
@@ -167,7 +167,6 @@ const YesNoContestResult = (props: {
   )
 
 interface Props {
-  ballotId?: string
   ballotStyleId: string
   election: Election
   isLiveMode: boolean
@@ -176,13 +175,13 @@ interface Props {
 }
 
 const PrintBallot = ({
-  ballotId = randomBase64(16),
   ballotStyleId,
   election,
   isLiveMode,
   precinctId,
   votes,
 }: Props) => {
+  const ballotId = randomBase64(16)
   const { county, date, seal, sealURL, state, parties, title } = election
   const partyPrimaryAdjective = getPartyPrimaryAdjectiveFromBallotStyle({
     ballotStyleId,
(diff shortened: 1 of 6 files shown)


src/utils/uuidSubstring.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Default is "xxxx-xxxx", where x is a hex char.
+ *
+ * @param {number} [indexStart=4] the start index
+ * @param {number} [indexEnd=9] the end index
+ *
+ * @return string
+ *
+ */
+
+const uuidSubstring = (indexStart: number = 4, indexEnd: number = 9) =>
```
eventualbuddha: This will _probably_ be fine, but it is definitely going to increase the chance of collisions. According to [Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier), the substring of a UUID with range `4...9` is the second half of `time_low` and all of `time_mid`, omitting `time_hi_and_version`, `clock_seq_hi_and_res`, `clock_seq_low`, and `node`. I'll defer to @benadida on the cryptographic significance of this.

benadida: thanks for the callout. Unfortunately, this won't work.

Chance of collision on a string is *much* higher than intuition. See the birthday paradox:
https://betterexplained.com/articles/understanding-the-birthday-paradox/

With 6 base-64 characters that's 36 bits, which means if those numbers were truly random, you'd get a 50% chance of a collision as soon as you have 2^18 ballots, aka 128,000 ballots. And with fewer ballots, still a healthy chance that it happens.

The accepted best practice to ensure no collisions ever is 128 bits, aka 16 bytes or 21 base-64 characters. I'd like to keep that for now.

beausmith: Thanks for the analysis gents. I'll break this into separate PRs for adding Ballot ID back and a second for a simpler, shorter, more human-friendly output.

PR #719, opened 2019-09-18
Stricter type check of ChoiceButton props

beausmith:


beausmith:
@benadida - This is related to, but not crucial to merge into the `try-pointer-events` PR.

PR #720, opened 2019-09-18
Bug: Tests Ballots were being hidden.

beausmith:
This was the bug seen on Tuesday where the test deck printed one blank page from VxPrint.

### DIFF
--- a/src/pages/TestBallotDeckScreen.tsx
+++ b/src/pages/TestBallotDeckScreen.tsx
@@ -122,10 +122,10 @@ const TestBallotDeckScreen = ({
 
   return (
     <React.Fragment>
-      <Main>
-        <MainChild maxWidth={false}>
-          {ballots.length ? (
-            <React.Fragment>
+      {ballots.length ? (
+        <React.Fragment>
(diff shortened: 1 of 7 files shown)


src/pages/TestBallotDeckScreen.tsx
```diff
@@ -122,10 +122,10 @@ const TestBallotDeckScreen = ({
 
   return (
     <React.Fragment>
-      <Main>
-        <MainChild maxWidth={false}>
-          {ballots.length ? (
-            <React.Fragment>
+      {ballots.length ? (
+        <React.Fragment>
```
eventualbuddha: Somewhat unrelated to this PR, but I just noticed it here. Do you have a preference between `<React.Fragment>…</React.Fragment>` and `<></>`?

beausmith: I do prefer `<React.Fragment>…</React.Fragment>` because it's more clear/obvious to me at a glance. But I do tend to prefer more concise syntax, so I could be convinced to use the shorter syntax.

PR #723, opened 2019-09-20
Landscape layout

beausmith:


src/AppContestMultiSeat.test.tsx
```diff
@@ -39,18 +39,14 @@ it('Single Seat Contest', async () => {
...
   currentCard = getNewVoterCard()
   advanceTimers()
 
-  // Go to Voting Instructions
-  await wait(() => fireEvent.click(getByText('Get Started')))
-  advanceTimers()
-
   // Go to First Contest
-  fireEvent.click(getByText('Start Voting'))
+  await wait(() => fireEvent.click(getAllByText('Start Voting')[1]))
```
benadida: the `[1]` is probably because it says "start voting" somewhere else... feels fragile for the test to work this way. Is there no way to say "look for a green button that says 'start voting?'

maybe something like the `getWithinKeyboard` trick you pulled below?

beausmith: Yes, it can be done another way but I figured that someone was gonna help me find a way to remove the text in the main part of the page. 

While this may be a little fragile, it's not fragile enough to worry about at this point.

src/components/Animations.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+//   91%, 93%, 95%, 97%, 99% {
+//     transform: translate3d(-0.3rem, 0, 0);
+//   }
+
+//   92%, 94%, 96%, 98% {
+//     transform: translate3d(0.3rem, 0, 0);
+//   }
+// `
+// animation: ${ shake2 } 10s ease-in-out infinite;
```
eventualbuddha: Remove commented out stuff?

src/lib/gamepad.test.tsx
```diff
@@ -42,22 +42,20 @@ it('gamepad controls work', async () => {
...
 
-  // for test coverage, we test pressing left and right on gamepad here, should do nothing
-  handleGamepadButtonDown('DPadLeft')
-  handleGamepadButtonDown('DPadRight')
-
-  // Go to Voting Instructions
-  fireEvent.click(getByText('Get Started'))
+  // // for test coverage, we test pressing left and right on gamepad here, should do nothing
+  // handleGamepadButtonDown('DPadLeft')
+  // handleGamepadButtonDown('DPadRight')
```
eventualbuddha: Intentionally commented out?

src/lib/gamepad.test.tsx
```diff
@@ -42,22 +42,20 @@ it('gamepad controls work', async () => {
...
 
   // Go to First Contest
-  advanceTimers()
-  fireEvent.click(getByText('Start Voting'))
+  handleGamepadButtonDown('DPadRight')
+  // advanceTimers()
```
eventualbuddha: Intentionally commented out?

src/lib/gamepad.test.tsx
```diff
@@ -76,7 +74,7 @@ it('gamepad controls work', async () => {
 
   // test the edge case of rolling over
   handleGamepadButtonDown('DPadUp')
-  expect(document.activeElement!.textContent).toEqual('Settings')
+  expect(document.activeElement!.textContent).toEqual('A') // Text Size Settings
```
eventualbuddha: Soon! https://github.com/microsoft/TypeScript/pull/33294

src/utils/ordinal.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Returns number param as ordinal.
+ * Source: https://stackoverflow.com/a/31615643/101290
+ *
+ * @param n number
+ *
+ * @return ordinal string
```
eventualbuddha: Rather than `@param` and `@return`, consider something like this:

```
@example

  ordinal(1)  // '1st'
  ordinal(42) // '42nd'
  ordinal(99) // '99th'
```

src/components/Animations.tsx
beausmith: Yeah. No one has complained about the current animation, so will remove.

src/lib/gamepad.test.tsx
beausmith: meant to delete. will do.

src/lib/gamepad.test.tsx
beausmith: meant to delete. no longer necessary due to removal of screen.

src/utils/ordinal.ts
beausmith: 👍🏼

src/lib/gamepad.test.tsx
beausmith: Yas!

beausmith:
@eventualbuddha @benadida - I've addressed all feedback with additional commits. Will merge to master once green.

PR #724, opened 2019-09-21
Write votes to long value, using the new b64 API

benadida:
Also, do the writes on an interval that waits for 1s after the last modification to write the data to the card.


src/AppRoot.tsx
```diff
@@ -165,6 +169,15 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
     }
   }
 
+  public fetchVotes = async () => {
+    const response = await fetch('/card/read_long_b64')
+    const returnValue = await response.json()
```
eventualbuddha: This could be `const returnValue = await fetchJSON('/card/read_long_b64')`, ideally with a type parameter for the return object shape.

src/AppRoot.tsx
```diff
@@ -254,19 +270,51 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
...
+      cardWriteInterval = window.setInterval(async () => {
+        if (
+          lastVoteSaveToCard < lastVoteUpdate &&
+          lastVoteUpdate < Date.now() - GLOBALS.CARD_WRITE_DELAY
+        ) {
+          lastVoteSaveToCard = Date.now()
+
+          const longValue = btoa(JSON.stringify(this.state.votes))
+
+          await this.storeLongValueOnCard(longValue)
```
eventualbuddha: Should we only call this if `longValue` is different from the last time?

eventualbuddha: In general, how do we want the server to know which write wins? In the event of requests `A` and `B` that are handled in the order `B` then `A`, how do we make sure `A` doesn't override `B`?

eventualbuddha: One possible solution: don't use `setInterval`, instead use `setTimeout` and only schedule a new timeout once the request has completed (or failed).

src/AppRoot.tsx
benadida: cool

src/AppRoot.tsx
benadida: @eventualbuddha good call on being extra safe about order of operations here. Will address.

benadida:
Ready for re-review.

src/AppRoot.tsx
benadida: I had trouble getting tests to pass with timers, and this seemed to be an additional complication, so I left it as is for now.

src/AppRoot.tsx
```diff
@@ -277,19 +294,60 @@ class AppRoot extends React.Component<RouteComponentProps, State> {
...
+        ) {
+          lastVoteSaveToCard = Date.now()
+
+          const longValue = btoa(JSON.stringify(this.state.votes))
+
+          writingVoteToCard = true
+          try {
+            await this.storeLongValueOnCard(longValue)
+          } catch (error) {
+            // eslint-disable-next-line no-empty
```
eventualbuddha: I'd replace this with an explanation for why we're ignoring the error.

src/AppEndToEnd.test.tsx
```diff
@@ -170,7 +173,14 @@ it('VxMark+Print end-to-end flow', async () => {
     // Vote for candidate contest
     if (title === presidentContest.title) {
       fireEvent.click(getByText(presidentContest.candidates[0].name))
+
+      // advance time by a second, run timers, and see if things get saved
+      advanceBy(1100)
```
eventualbuddha: The comment is out of sync with the code. Replace "advance time by a second" with the reason we need to wait, i.e. what are we waiting for?

package.json
```diff
@@ -112,6 +112,7 @@
     "@typescript-eslint/eslint-plugin": "1.13.0",
     "@typescript-eslint/parser": "^1.13.0",
     "cypress": "^3.4.1",
+    "jest-date-mock": "^1.0.7",
```
eventualbuddha: Should we add this to `setupFiles` per the [README](https://github.com/hustcc/jest-date-mock#setup)?

src/AppRoot.tsx
```diff
@@ -129,6 +129,11 @@ const initialState: State = {
 
 let checkCardInterval = 0
 
+let lastVoteUpdate = 0
+let lastVoteSaveToCard = 0
+let cardWriteInterval = 0
+let writingVoteToCard = false
```
eventualbuddha: These should all be private instance variables on `AppRoot`, IMO. Otherwise each instance will share these values. This happens to not be a problem because–in testing–jest loads this class multiple times and resets the state, and in real use this component is never unmounted.

That said, it's just waiting to bite us when we do this somewhere that doesn't have this property. Can this functionality be split out into its own class? This file is getting awfully big.

package.json
benadida: I tried but that requires ejecting from create-react-app and it looked unnecessary for what we're doing.

src/AppEndToEnd.test.tsx
```diff
@@ -170,7 +173,14 @@ it('VxMark+Print end-to-end flow', async () => {
...
+      advanceTimers()
+
+      expect(fetchMock.calls('/card/write_long_b64')).toHaveLength(1)
```
beausmith: There should be tests to advance time and expect no change in fetchMock.calls too.

src/AppEndToEnd.test.tsx
benadida: I don't think it's out of sync, but I'll clarify the comment.

test/helpers/smartcards.ts
```diff
@@ -45,9 +45,11 @@ export const pollWorkerCard: CardPresentAPI = {
 }
 
 export const createVoterCard = (
-  config?: Partial<VoterCardData>
+  config?: Partial<VoterCardData>,
+  longValueExists?: boolean
```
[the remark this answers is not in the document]
benadida: I hear you, but this is used with an extra param in only one place, and changing this makes the rest of the code that doesn't use the second parameter a good bit uglier. I'd like to keep it as is for now.

src/AppEndToEnd.test.tsx
```diff
@@ -167,7 +170,22 @@ it('VxMark+Print end-to-end flow', async () => {
...
+
+      // We write to the card when no changes to the ballot state have happened for a second.
+      // To test that this is happening, we advance time by a bit more than a second
+      // We also need to advance timers so the interval will run, see that time has passed,
+      // and finally write to the card.
+      advanceBy(1100)
+      advanceTimers()
+      expect(fetchMock.calls('/card/write_long_b64')).toHaveLength(1)
+
+      // If we wait another second and avance timers, without any change made to the card,
```
eventualbuddha:
```suggestion
      // If we wait another second and advance timers, without any change made to the card,
```

test/helpers/smartcards.ts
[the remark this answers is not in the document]
eventualbuddha: It could be added to `config`. Something like this:

```ts
type CreateVoterCardConfig = Partial<VoterCardData> & {
  longValueExists?: boolean
}

export const createVoterCard = (
  config: CreateVoterCardConfig = {}
): CardPresentAPI => {
  const { longValueExists, ...cardData } = config
  return {
    present: true,
    longValueExists,
    shortValue: JSON.stringify({
      t: 'voter',
      c: utcTimestamp(),
      pr: election.precincts[0].id,
      bs: election.ballotStyles[0].id,
      ...cardData,
    }),
  }
}
```

src/AppRoot.tsx
beausmith: Addressing this: https://github.com/votingworks/bmd/compare/master...catch-long-value-write?expand=1#files_bucket

PR #731, opened 2019-09-24
Feedback on landscape flow

beausmith:


src/AppContestMultiSeat.test.tsx
```diff
@@ -79,12 +79,10 @@ it('Single Seat Contest', async () => {
   expect(container.firstChild).toMatchSnapshot()
 
   // Go to Review Screen
-  while (!queryByText('Review →')) {
+  while (!queryByText('All Your Votes')) {
```
benadida: does it still say "All your votes" now? Should it say all your choices? Or is this appearing elsewhere?

beausmith: Here is a screenshot where you can see all this language:

image

beausmith: Decided to revert to using "votes" and be sensitive to see if any voters on the review screen believe that they are casting more votes than they are supposed to.

PR #736, opened 2019-09-24
Increase expired time and idle timeout.

### LINKED ISSUE #732: Increase expired card to 60 minutes.

### LINKED ISSUE #735: Increase idle timeout to 5 min and "I'm still voting" after 1 minute, then invalidate card.

beausmith:
Fixes #732
Fixes #735


beausmith:
@eventualbuddha - Might be worth a post-merge review?

src/components/Animations.tsx
```diff
@@ -12,97 +12,4 @@ export const Wobble = styled.div`
...
-//   91%, 93%, 95%, 97%, 99% {
-//     transform: translate3d(-0.3rem, 0, 0);
-//   }
-
-//   92%, 94%, 96%, 98% {
-//     transform: translate3d(0.3rem, 0, 0);
-//   }
-// `
-// animation: ${ shake2 } 10s ease-in-out infinite;
+export default {}
```
eventualbuddha: Can we just disable the eslint rule to prefer default export? I don't think it's being very helpful.

src/pages/IdlePage.tsx
```diff
@@ -37,7 +38,7 @@ const IdlePage = () => {
   }, [countdown, markVoterCardVoided, resetBallot])
 
   useInterval(() => {
-    setCountdown(t => (t > 0 ? t - 1 : 0))
+    setCountdown(countdown => countdown - 1)
```
eventualbuddha: For this use case I don't think this matters much, but there's no guarantee that `setInterval` and `setTimeout` will actually run "on time". It's better to have an interval to refresh the display, but to make the data be based on an absolute timestamp, i.e. we calculate the idle timeout time upfront and update the display based on that.

PR #753, opened 2019-09-26
seal optimized

beausmith:


beausmith:
@benadida - I don't understand why the cypress test is not passing. I can run these tests locally and they all pass.

PR #755, opened 2019-09-26
Open/Close Polls Report is one action.

### LINKED ISSUE #743: Print - reduce print open/close polls report to 1 action - OK to fail when no module-print.

beausmith:
fixes #743 

Test coverage is not at 100%, but you'll see the uncovered bits are non-critical. Would like to have coverage, but willing to reduce coverage temporarily.


benadida:
I'm fine with test coverage exceptions for these conditions. They are super edge cases that I'm not worried about.

benadida:
filing a ticket to review the test coverage skips. #760

PR #764, opened 2019-09-28
feat: encode ballot data for QR code with `ballot-encoder`

eventualbuddha:
This changes from a text-based format to a compact binary format. See https://github.com/votingworks/ballot-encoder/pull/1 and https://github.com/votingworks/module-scan/pull/31.


src/pages/__snapshots__/PrintPage.test.tsx.snap
```diff
@@ -1,486 +1,591 @@
...
-  <main
+  <div
     class="c1"
   >
     <div
-      class="c2"
+      class="c2 ballot-header-content"
+    >
+      <h2>
+        Unofficial TEST Ballot
```
eventualbuddha: @beausmith it appears these snapshots were only capturing the "Printing official ballot" content and not the ballot itself. I fixed that by referring to `childNodes[1]` instead of `firstChild`.

beausmith: 👍

PR #766, opened 2019-09-28
assume backend unavailable and reset

beausmith:
To address this: https://github.com/votingworks/bmd/pull/724/files#r328712166

Needs Tests!


benadida:
I'm not sure we want to be that aggressive. Could be a transient failure of writing, and the goal of the catch is to ensure we always survive a transient failure in writing to the card.

PR #780, opened 2019-10-01
gate printing so there's only one print per card insert

benadida:


src/pages/PrintOnlyScreen.tsx
```diff
@@ -58,29 +67,28 @@ const PrintOnlyScreen = ({
...
     }
-  }, [cardVotes, printBallot])
+  }, [cardVotes, printBallot, isReadyToPrint, okToPrint, setOkToPrint])
 
   useEffect(() => {
     if (!isVoterCardPresent) {
       updateIsPrinted(false)
+
+      // once card is taken out, ok to print again
+      if (!okToPrint) {
```
beausmith: Is this conditional check is unnecessary. Setting state to what it already is does not cause a re-render.

src/pages/PrintOnlyScreen.tsx
```diff
@@ -58,29 +67,28 @@ const PrintOnlyScreen = ({
...
       printBallot()
     }
-  }, [cardVotes, printBallot])
+  }, [cardVotes, printBallot, isReadyToPrint, okToPrint, setOkToPrint])
 
   useEffect(() => {
     if (!isVoterCardPresent) {
       updateIsPrinted(false)
+
+      // once card is taken out, ok to print again
```
beausmith: nit: Unnecessary comment. Remove.

PR #783, opened 2019-10-02
single state change for voter card

benadida:


src/AppRefresh.test.tsx
```diff
@@ -63,6 +76,10 @@ it('Refresh window and expect to be on same contest', async () => {
   advanceTimers()
   expect(getByText(candidate0).closest('button')!.dataset.selected).toBe('true')
 
+  // advance time to give opportunity for background interval to write to card
+  advanceBy(1100)
```
beausmith: I think we should clearly identify how numbers are calculated, otherwise the number borders on ✨ magic ✨ .

```suggestion
  // Advance time by CARD_LONG_VALUE_WRITE_DELAY + 100ms to give opportunity for background interval to write to card.
  advanceBy(1000 + 100)
```

beausmith: Also… thanks for figuring out this issue with code wanting to execute after unmount. 👍

PR #803, opened 2019-10-07
trying a new approach to touch that allows for swiping pan

benadida:
This works, though it doesn't feel quite as fast as taking action on `touchstart`.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -62,15 +62,16 @@ const Button = ({
   const [startCoordinates, setStartCoordinates] = useState([0, 0])
 
   const onTouchStart = (event: React.TouchEvent) => {
-    const touch = event.touches[0]
-    setStartCoordinates([touch.clientX, touch.clientY])
+    const { clientX, clientY } = event.touches[0]
+    setStartCoordinates([clientX, clientY])
   }
 
   const onTouchEnd = (event: React.TouchEvent) => {
-    const touch = event.changedTouches[0]
+    const maxMove = 30
+    const { clientX, clientY } = event.changedTouches[0]
     if (
-      Math.abs(startCoordinates[0] - touch.clientX) < 30 &&
-      Math.abs(startCoordinates[1] - touch.clientY) < 30
+      Math.abs(startCoordinates[0] - clientX) < maxMove &&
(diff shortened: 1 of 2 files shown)


eventualbuddha:
This does work, though it makes me wonder even more strongly about why the default behavior isn't doing what we want. Have you played around with https://rbyers.github.io/eventTest.html to see exactly what events are happening when you do the click and scroll behaviors you're testing?

benadida:
@eventualbuddha I haven't played with that specifically. I will try. I can tell you that the events I'm seeing are: touchstart, touchmove, touchend in all cases. For pointer events, a bit too much moving causes a pointercancel, thus no pointerup.

src/components/Button.tsx
```diff
@@ -37,7 +37,7 @@ const buttonStyles = css<StyledButtonProps>`
     (primary && '#FFFFFF') ||
     'black'};
   font-size: ${({ big = false }) => (big ? '1.25rem' : undefined)};
-  touch-action: manipulation;
+  touch-action: none;
```
beausmith: Have you tried removing this line? Is this line necessary?

benadida: same either way, I can remove it.

benadida:
@eventualbuddha @beausmith no matter what I try, I'm not seeing how to simplify this. I didn't learn anything from that resource @eventualbuddha.

What do you expect to see happen? It's not clear to me that there is an obvious default behavior for the browser.

src/components/Button.tsx
```diff
@@ -59,24 +59,30 @@ const Button = ({
...
-      onPress(event)
-    } else {
-      setReadyForEvents(true)
+  const onTouchEnd = (event: React.TouchEvent) => {
+    const touch = event.changedTouches[0]
+    if (
+      Math.abs(startCoordinates[0] - touch.clientX) < 30 &&
+      Math.abs(startCoordinates[1] - touch.clientY) < 30
+    ) {
+      onPress(event as any)
```
eventualbuddha: I wonder whether we should change `onPress`’s signature to make this kind of casting not necessary. Looks like we only use `currentTarget`, maybe? Not needed for this PR, just been thinking about this.

PR #809, opened 2019-10-08
Larger type and graphics

### LINKED ISSUE #801: Bigger Text & Graphics on VxMark & VxPrint
## VxEncode
- [x] VxEncode may size graphics by % of screen, not by pixels. (https://github.com/votingworks/bas/pull/55)

## VxPrint
- [x] Bigger graphics on VxPrint. 
- [x] VxPrint may size graphics by % of screen, not by pixels.
- [x] VxPrint Pictograms and instructions on are way too small, make them as big as you can.
- [x] VxPrint polls/open close on has text and button that are too small on high-res Surface Go. 

## VxMark
- [x] Bigger graphics on VxMark,
- [x] VxMark graphics & instructions are too small on Lenovo 1080p , make them as big as possible.

beausmith:
Fixes #801 

Definitely give this one a test drive. It's fun to drive. 😄


eventualbuddha:
On the first sample contest (President), the "See More" buttons scroll down/up just barely short of what would make it no longer scroll in that direction. That is, pressing "See More" to scroll down scrolls "Martin Patterson" into view and leaves the "See More" button enabled. Pressing it again scrolls a minute amount before getting to the end. Here are the relevant values from the end of `CandidateContest#scrollContestChoices` when hitting down twice then up twice:

image

Maybe we should clamp it to max/min if it would have < ~44 left?

eventualbuddha:
image

"See More" needs a z-index adjustment.

eventualbuddha:
Does the VxMark+Print mode want graphics?

image

src/AppContestYesNo.test.tsx
```diff
@@ -39,29 +39,28 @@ it('Single Seat Contest', async () => {
...
-  await wait(() => fireEvent.click(getAllByText('Start Voting')[1]))
+  await wait(() => fireEvent.click(getByText('Start Voting')))
   advanceTimers()
 
   // ====================== END CONTEST SETUP ====================== //
 
   // Query by text which includes markup.
   // https://stackoverflow.com/questions/55509875/how-to-query-by-text-string-which-contains-html-tags-using-react-testing-library
-  const getByTextWithMarkup = (text: string) => {
+  const getByTextWithMarkup = (text: string): HTMLElement =>
```
eventualbuddha: Time to pull this into a helper?

src/components/SettingsTextSize.tsx
```diff
@@ -55,7 +55,7 @@ const SettingsTextSize = ({ userSettings, setUserSettings }: Props) => {
           <Label>Change Text Size</Label>
           <FontSizeButtons>
             <SegmentedButton data-testid="change-text-size-buttons">
-              {FONT_SIZES.map((v: number, i: number) => (
+              {FONT_SIZES.slice(0, 3).map((v: number, i: number) => (
```
eventualbuddha: There are four font sizes, but we can only access three of them?

src/pages/ClerkScreen.tsx
```diff
@@ -96,11 +99,15 @@ const ClerkScreen = ({
...
                       Select a precinct for this device…
                     </option>
-                    {election.precincts.map(precinct => (
-                      <option key={precinct.id} value={precinct.id}>
-                        {precinct.name}
-                      </option>
-                    ))}
+                    {election.precincts
+                      .sort((a, b) =>
+                        alphaSort(a.name.toLowerCase(), b.name.toLowerCase())
```
eventualbuddha: This is better done using [`String#localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) like so:

```ts
.sort((a, b) =>
  a.name.localeCompare(b.name, undefined, {
    ignorePunctuation: true,
  })
)
```

src/pages/ExpiredCardScreen.tsx
```diff
@@ -1,19 +1,33 @@
...
+  setUserSettings: (partial: PartialUserSettings) => void
+}
+
+const ExpiredCardScreen = ({ setUserSettings }: Props) => {
+  useEffect(() => {
+    setUserSettings({ textSize: 3 })
+    return () => {
+      setUserSettings({ textSize: 1 })
+    }
+  }, [setUserSettings])
```
eventualbuddha: This is a little odd. I guess it doesn't matter that you've lost the voter's settings because they can't continue anyway. Should the default of 1 come from a constant?

src/pages/PrintOnlyScreen.tsx
```diff
@@ -86,8 +93,12 @@ const PrintOnlyScreen = ({
   }, [isVoterCardPresent, okToPrint, setOkToPrint])
 
   useEffect(() => {
-    return () => clearTimeout(printerTimer.current)
-  }, [])
+    setUserSettings({ textSize: 3 })
+    return () => {
+      setUserSettings({ textSize: 1 })
```
eventualbuddha: Ditto on default.

src/pages/UsedCardScreen.tsx
```diff
@@ -1,19 +1,33 @@
...
+  setUserSettings: (partial: PartialUserSettings) => void
+}
+
+const UsedCardScreen = ({ setUserSettings }: Props) => {
+  useEffect(() => {
+    setUserSettings({ textSize: 3 })
+    return () => {
+      setUserSettings({ textSize: 1 })
+    }
+  }, [setUserSettings])
```
eventualbuddha: Should this whole thing be a custom `useX`?

beausmith:
Responding to @eventualbuddha comments…

> scroll min/max

Def want to to. Making issue: https://github.com/votingworks/bmd/issues/811

> z-index issue

Fixed.

> "Does the VxMark+Print mode want graphics?" 

Yes, later.

src/AppContestYesNo.test.tsx
beausmith: Yep. Will do.

src/components/SettingsTextSize.tsx
beausmith: Yes. This is to allow the VxPrint UI to use the same `SettingsTextSize` mechanism to display larger content. Eventually we should replace this simple changing of the root `font-size` with "themes".

src/pages/ClerkScreen.tsx
beausmith: Great!

src/pages/ExpiredCardScreen.tsx
beausmith: will do.

src/pages/PrintOnlyScreen.tsx
beausmith: will do.

src/pages/UsedCardScreen.tsx
beausmith: Yeah… that might be better. I'd like to put the effort into themes rather than further support this clever hack. 😝

PR #815, opened 2019-10-09
Use arrow icons for next and back buttons.

beausmith:
image

image


beausmith:
Changed arrow from solid to stroked style to make them less heavy.

PR #836, opened 2019-10-14
feat: print PDFs via module-print if possible

eventualbuddha:
Requires https://github.com/votingworks/module-print/pull/22


eventualbuddha:
I've mostly just let the tools do it. VS Code automatically adds imports somewhere. `eslint-plugin-import` has some preferences about it, so I let it adjust the changes VS Code makes. If it looks like it's becoming a mess, I just hit `⇧⌥O` (`Organize Imports`).

PR #949, opened 2019-12-04
fix: limit write-in candidate names to 40 chars

### LINKED ISSUE #849: Enforce write-in limit in modal
Enforce Write-in length to 40 characters. Disable all keys, except "delete" key. 

@mcchilders Says: if you overflow the character limit on write-ins it warns you but doesn’t stop you from continuing to enter characters even after the limit, which borks both the UI and causes the ballot to print with a blank vote in that contest instead of your write-in. Pretty sure it should enforce the limit it’s warning you about?

eventualbuddha:
We had a soft limit where we would show a warning, now we have a hard limit that actually stops input.

Fixes #849

write-in-limit


src/components/CandidateContest.test.tsx
```diff
@@ -137,4 +164,44 @@ describe('supports write-in candidates', () => {
...
+      />
+    )
+    fireEvent.click(getByText('add write-in candidate').closest('button')!)
+    getByText('Write-In Candidate')
+    const writeInCandidate =
+      "JACOB JOHANSON JINGLEHEIMMER SCHMIDTT, THAT'S MY NAME TOO"
+    Array.from(writeInCandidate).forEach(i => {
+      const key = i === ' ' ? 'space' : i
+      fireEvent.click(getByText(key).closest('button')!)
+    })
```
beausmith: Should this function be abstracted as its used more than once?

src/components/CandidateContest.test.tsx
```diff
@@ -137,4 +164,44 @@ describe('supports write-in candidates', () => {
...
+    getByText('You have entered 40 of maximum 40 characters.')
+
+    expect(
+      getByText('space')
+        .closest('button')!
+        .hasAttribute('disabled')
+    ).toBe(true)
+    expect(container).toMatchSnapshot()
```
beausmith: delete?

src/config/globals.ts
```diff
@@ -13,3 +13,4 @@ export enum YES_NO_VOTES {
   no = 'No',
   yes = 'Yes',
 }
+export const MAX_WRITE_IN_CANDIDATE_LENGTH = 40
```
beausmith: nit: 
```suggestion
export const WRITE_IN_CANDIDATE_MAX_LENGTH = 40
```

src/components/CandidateContest.tsx
```diff
@@ -639,7 +652,10 @@ class CandidateContest extends React.Component<Props, State> {
                     <WriteInCandidateCursor />
                   </WriteInCandidateName>
                 </WriteInCandidateFieldSet>
-                <VirtualKeyboard onKeyPress={this.onKeyboardInput} />
+                <VirtualKeyboard
+                  onKeyPress={this.onKeyboardInput}
+                  keyDisabled={this.keyDisabled}
```
beausmith: nice!

beausmith:
Thanks for posting the animated gif of this in action. Very helpful.

src/components/CandidateContest.test.tsx
```diff
@@ -114,6 +114,33 @@ describe('supports multi-seat contests', () => {
...
+        updateVote={updateVote}
+      />
+    )
+    fireEvent.click(getByText('add write-in candidate').closest('button')!)
+    getByText('Write-In Candidate')
+    Array.from('LIZARD PEOPLE').forEach(i => {
+      const key = i === ' ' ? 'space' : i
+      fireEvent.click(getByText(key).closest('button')!)
+    })
+    expect(container).toMatchSnapshot()
```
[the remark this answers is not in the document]
eventualbuddha: This was mostly cargo-culted from what was here before. I think it's reasonable to say we can remove it given the assertions I added.

PR #961, opened 2019-12-10
[WIP] ElectionGuard

benadida:
This has hard-wired configuration, isn't ready for merging.


eventualbuddha:
There's no feature flagging built into this. Is that the plan, or do we keep it on a separate branch?

benadida:
> There's no feature flagging built into this. Is that the plan, or do we keep it on a separate branch?

Plan is to have the flag + configuration info in the election definition. Not merging until we have that done.

PR #966, opened 2019-12-12
Built-in screen reader

eventualbuddha:
This PR has a very very VERY basic screen reader built into BMD.

**Update 12/13:** By default the screen reader is off. To turn it on, press `alt+r`.
**Update 12/17:** By default the screen reader is off. To turn it on, press `r`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -90,5 +90,6 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off', // Want to use it, but it requires return types for all built-in React lifecycle methods.
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-use-before-define': ['error', { functions: false }],
+    'import/extensions': 'off'
   },
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useCallback } from 'react'
 import { BrowserRouter, Route } from 'react-router-dom'
 
 import 'normalize.css'
@@ -7,18 +7,111 @@ import './App.css'
 import FocusManager from './components/FocusManager'
 
 import AppRoot from './AppRoot'
+import {
+  ScreenReader,
+  AriaScreenReader,
+  SpeechSynthesisTextToSpeech,
+  NullTextToSpeech,
+  TextToSpeech,
+} from './utils/ScreenReader'
 
 /* istanbul ignore next - unsure how to test */
 window.oncontextmenu = (e: MouseEvent): void => {
   e.preventDefault()
 }
 
-const App = () => (
-  <BrowserRouter>
-    <FocusManager>
-      <Route path="/" component={AppRoot} />
-    </FocusManager>
-  </BrowserRouter>
-)
+export interface Props {
+  tts?: {
+    enabled: TextToSpeech
+    disabled: TextToSpeech
+  }
+}
+
+const App = ({
+  tts = {
+    enabled: new SpeechSynthesisTextToSpeech(),
+    disabled: new NullTextToSpeech(),
+  },
+}: Props) => {
+  const [screenReaderEnabled, setScreenReaderEnabled] = useState(false)
+  const [screenReader, setScreenReader] = useState<ScreenReader>(
+    new AriaScreenReader(tts.disabled)
+  )
+
+  /* istanbul ignore next - need to figure out how to test this */
+  const onKeyPress = useCallback(
+    (event: React.KeyboardEvent) => {
+      if (event.key === 'r') {
+        if (screenReaderEnabled) {
+          screenReader.onScreenReaderDisabled()
+          setScreenReader(new AriaScreenReader(tts.disabled))
+          setScreenReaderEnabled(false)
+        } else {
+          const newScreenReader = new AriaScreenReader(tts.enabled)
+          setScreenReader(newScreenReader)
+          setScreenReaderEnabled(true)
+          newScreenReader.onScreenReaderEnabled()
+        }
+      }
+    },
+    [
+      screenReader,
+      setScreenReader,
+      screenReaderEnabled,
+      setScreenReaderEnabled,
+      tts.disabled,
+      tts.enabled,
+    ]
+  )
+
+  /* istanbul ignore next - need to figure out how to test this */
+  const onClick = useCallback(
+    ({ target }: React.MouseEvent) => {
+      if (target) {
+        const currentPath = window.location.pathname
+
+        setImmediate(() => {
+          // Only send `onClick` to the screen reader if the click didn't
+          // trigger navigation.
+          if (window.location.pathname === currentPath) {
+            screenReader.onClick(target)
+          }
+        })
+      }
+    },
+    [screenReader]
+  )
+
+  /* istanbul ignore next - need to figure out how to test this */
+  const onFocus = useCallback(
+    ({ target }: React.FocusEvent) => {
+      if (target) {
+        const currentPath = window.location.pathname
+
+        setImmediate(() => {
+          // Only send `onFocus` to the screen reader if the focus didn't
+          // trigger navigation.
+          if (window.location.pathname === currentPath) {
+            screenReader.onFocus(target)
+          }
+        })
+      }
+    },
+    [screenReader]
+  )
+
+  return (
+    <BrowserRouter>
+      <FocusManager
+        screenReader={screenReader}
+        onKeyPress={onKeyPress}
+        onClickCapture={onClick}
+        onFocusCapture={onFocus}
+      >
+        <Route path="/" component={AppRoot} />
+      </FocusManager>
+    </BrowserRouter>
+  )
+}
 
 export default App
--- a/src/components/LinkButton.tsx
+++ b/src/components/LinkButton.tsx
@@ -32,15 +32,9 @@ const LinkButton = (props: Props) => {
     if (onPress) {
       onPress(event)
     } else if (goBack && !to) {
-      // Delay to avoid passing tap to next screen
-      window.setTimeout(() => {
-        history.goBack()
-      }, 100)
+      history.goBack()
     } else if (to && !goBack) {
-      // Delay to avoid passing tap to next screen
-      window.setTimeout(() => {
-        history.push(to)
-      }, 100)
+      history.push(to)
     }
   }
   return (
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -54,7 +54,8 @@ const Modal: React.FC<Props> = ({
       document.getElementById('root')! || document.body.firstElementChild
     }
     ariaHideApp
-    role="none"
+    aria-modal
+    role="alertdialog"
     isOpen={isOpen}
     contentLabel={ariaLabel}
     portalClassName="modal-portal"
--- a/src/components/YesNoContest.tsx
+++ b/src/components/YesNoContest.tsx
@@ -361,7 +361,7 @@ export default class YesNoContest extends React.Component<Props> {
           content={
             <Prose>
               {overvoteSelection && (
-                <p>
+                <p id="modalaudiofocus">
                   Do you want to change your vote to{' '}
                   <strong>{YES_NO_VOTES[overvoteSelection]}</strong>? To change
                   your vote, first unselect your vote for{' '}
--- a/src/lib/gamepad.test.tsx
+++ b/src/lib/gamepad.test.tsx
@@ -50,7 +50,6 @@ it('gamepad controls work', async () => {
 
   // Go to First Contest
   handleGamepadButtonDown('DPadRight')
-  fireEvent.click(getByText('Start Voting'))
   advanceTimers()
 
   // First Contest Page
--- a/src/lib/gamepad.ts
+++ b/src/lib/gamepad.ts
@@ -1,8 +1,7 @@
 import { Button } from 'react-gamepad'
 import mod from '../utils/mod'
 
-export const getActiveElement = () =>
-  document.activeElement! as HTMLInputElement
+export const getActiveElement = () => document.activeElement! as HTMLElement
 
 function getFocusableElements(): HTMLElement[] {
   const tabbableElements = Array.from(
@@ -41,7 +40,7 @@ function handleArrowDown() {
 }
 
 function handleArrowLeft() {
-  const prevButton = document.getElementById('previous') as HTMLButtonElement
+  const prevButton = document.getElementById('previous')
   /* istanbul ignore else */
   if (prevButton) {
     prevButton.click()
@@ -49,7 +48,7 @@ function handleArrowLeft() {
 }
 
 function handleArrowRight() {
-  const nextButton = document.getElementById('next') as HTMLButtonElement
+  const nextButton = document.getElementById('next')
   /* istanbul ignore else */
   if (nextButton) {
     nextButton.click()
--- a/src/pages/ContestPage.tsx
+++ b/src/pages/ContestPage.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react'
+import React, { useContext, useRef } from 'react'
 import { Redirect, RouteComponentProps } from 'react-router-dom'
 import { CandidateVote, OptionalYesNoVote } from '@votingworks/ballot-encoder'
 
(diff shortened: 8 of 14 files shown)


benadida:
Other than the audio triggering briefly when you select a button (which @eventualbuddha already mentioned via voice), this is working very well and is on-par with what we get from chromevox. Nicely done!

eventualbuddha:
@benadida I fixed the bug where it started to say the button e.g. "Next" before moving on to read the newly-navigated-to page. However, there's still the same problem with modals. I think it's something we should file in an issue and come back to.

eventualbuddha:
@beausmith I got a little trigger-happy on this and forgot that I wanted your review. Mind taking a look? Since it's already merged, I'll make new issues from your comments.

src/components/LinkButton.tsx
```diff
@@ -32,15 +32,9 @@ const LinkButton = (props: Props) => {
...
-      // Delay to avoid passing tap to next screen
-      window.setTimeout(() => {
-        history.goBack()
-      }, 100)
+      history.goBack()
     } else if (to && !goBack) {
-      // Delay to avoid passing tap to next screen
-      window.setTimeout(() => {
-        history.push(to)
-      }, 100)
```
eventualbuddha: These timeouts did not seem to be required anymore. I tried it both with a mouse and with a touchscreen and I never saw clicks "bleed through". I removed them because they were making the screen reader's job much harder.

src/components/Modal.tsx
```diff
@@ -54,7 +54,8 @@ const Modal: React.FC<Props> = ({
       document.getElementById('root')! || document.body.firstElementChild
     }
     ariaHideApp
-    role="none"
+    aria-modal
+    role="alertdialog"
```
eventualbuddha: These seemed to be appropriate attribute values for a modal, though nothing in this PR actually requires them.

src/components/YesNoContest.tsx
```diff
@@ -361,7 +361,7 @@ export default class YesNoContest extends React.Component<Props> {
           content={
             <Prose>
               {overvoteSelection && (
-                <p>
+                <p id="modalaudiofocus">
```
eventualbuddha: This is carried over from `CandidateContest`.

src/lib/gamepad.test.tsx
```diff
@@ -50,7 +50,6 @@ it('gamepad controls work', async () => {
 
   // Go to First Contest
   handleGamepadButtonDown('DPadRight')
-  fireEvent.click(getByText('Start Voting'))
```
eventualbuddha: This was no longer needed because I removed the `LinkButton` activation delay.

src/lib/gamepad.ts
```diff
@@ -1,8 +1,7 @@
 import { Button } from 'react-gamepad'
 import mod from '../utils/mod'
 
-export const getActiveElement = () =>
-  document.activeElement! as HTMLInputElement
```
eventualbuddha: Nothing was relying on this being an `HTMLInputElement`, and it typically won't be, so I just made it an `HTMLElement`.

src/pages/ContestPage.tsx
```diff
@@ -58,7 +59,7 @@ const ContestPage = (props: RouteComponentProps<ContestParams>) => {
   // - confirm intent when navigating away without selecting a candidate
 
   return (
-    <Screen>
+    <Screen ref={rootRef as any}>
```
eventualbuddha: All changes in this file are unnecessary and represent a path I did not end up taking. I was trying to get a reference to the root element of the screen for the purpose of adding/removing event handlers, but I ended up doing that by adding handlers to `FocusManager`. I'll remove these.

src/setupTests.tsx
```diff
@@ -25,6 +25,39 @@ window.print = jest.fn(() => {
...
+    dispatchEvent: jest.fn(),
+    getVoices: jest.fn(),
+    onvoiceschanged: jest.fn(),
+    pause: jest.fn(),
+    paused: false,
+    pending: false,
+    removeEventListener: jest.fn(),
+    resume: jest.fn(),
+    speak: jest.fn(),
+    speaking: false,
```
eventualbuddha: It'd be great if all of this was just part of `jsdom`, but [a search for `speechSynthesis` in that repo literally returns nothing](https://github.com/jsdom/jsdom/search?q=speechsynthesis&unscoped_q=speechsynthesis).

src/utils/ScreenReader.test.ts
```diff
@@ -0,0 +1,353 @@
...
+  for (const key in attributes) {
+    if (Object.prototype.hasOwnProperty.call(attributes, key)) {
+      element.setAttribute(key, attributes[key])
+    }
+  }
+
+  element.append(...children)
+
+  return element
+}
```
eventualbuddha: This helper should probably be in `testUtils`.

src/utils/ScreenReader.test.ts
```diff
@@ -0,0 +1,353 @@
...
+    const tts = new SpeechSynthesisTextToSpeech()
+
+    tts.speak('hello right now!', { now: true })
+    expect(speechSynthesis.cancel).toHaveBeenCalled()
+    expect(speechSynthesis.speak).toHaveBeenCalledWith(
+      expect.objectContaining({ text: 'hello right now!' })
+    )
+  })
+
+  it('resolves when onend is called', async () => {
```
eventualbuddha: This test ended up being tricky to write. Any suggestions for making it clearer/simpler would be appreciated.

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+   * @param tts A text-to-speech engine to use to speak aloud.
+   */
+  public constructor(tts: TextToSpeech) {
+    this.tts = tts
+  }
+
+  /**
+   * Call this with an event target when a focus event occurs. Resolves when speaking is done.
+   */
+  public onFocus = async (target?: EventTarget) => {
```
eventualbuddha: This and `onClick` don't need to be bound functions.

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+  public async onScreenReaderDisabled(): Promise<void> {
+    await this.speak('Screen reader disabled', { now: true })
+  }
+
+  /**
+   * Directly triggers speech of text. Resolves when speaking is done.
+   */
+  public async speak(text: string, options: SpeakOptions = {}): Promise<void> {
+    /* istanbul ignore next */
+    if (process.env.NODE_ENV === 'development') {
```
eventualbuddha: This is here so we don't get a bunch of junk when the tests run, but it's useful to see in development.

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+    return description
+      .replace(/ {2}/g, ' ')
+      .replace(/^[. ]+/g, '')
+      .replace(/\. +\./g, '.')
+      .replace(/,\./g, '.')
+      .replace(/ +\./g, '.')
+      .replace(/ +,/g, ',')
+      .replace(/\.+/g, '.')
+      .replace(/ +$/g, '')
+      .replace(/^ +/g, '')
```
eventualbuddha: In practice I found this to not be strictly necessary from a text-to-speech perspective. The systems I tried this with all did fine even when there were a bunch of consecutive periods or a bunch of spaces or random comma-period pairs. However, when logging the spoken text it's definitely nicer to have something sensible so this just cleans it up.

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+    return (
+      Array.from(node.childNodes)
+        .map(child => this.describe(child))
+        .filter(Boolean)
+        .join(' ') + terminator
+    )
+  }
+
+  /**
+   * Determines wheether `element` is a block or inline element.
```
eventualbuddha: Typo: "whether"

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+    if (
+      element.hasAttribute('aria-hidden') &&
+      element.getAttribute('aria-hidden') !== 'false'
+    ) {
+      return true
+    }
+
+    const style = getComputedStyle(element)
+
+    if (style.display === 'none' || style.visibility === 'hidden') {
```
eventualbuddha: It's possible there are more ways to hide an element than this, but these were the obvious ones.

src/components/Modal.tsx
beausmith: If not required or not useful, wouldn't it be better to delete 'em?

src/setupTests.tsx
beausmith: It's lonely on the bleeding edge of A11y. 🤦🏼‍♂️

src/utils/ScreenReader.ts
beausmith: 👍🏼 I promise not to hide elements using any other method. Eventually there may be elements animating from off-screen to on-screen, but the screen reader should be able to treat elements the same way as they do now.

beausmith:
@eventualbuddha - Sorry for delay. Reviewing now…

src/App.tsx
```diff
@@ -7,18 +7,111 @@ import './App.css'
...
+    enabled: new SpeechSynthesisTextToSpeech(),
+    disabled: new NullTextToSpeech(),
+  },
+}: Props) => {
+  const [screenReaderEnabled, setScreenReaderEnabled] = useState(false)
+  const [screenReader, setScreenReader] = useState<ScreenReader>(
+    new AriaScreenReader(tts.disabled)
+  )
+
+  /* istanbul ignore next - need to figure out how to test this */
```
beausmith: I'm having flashbacks from trying to test the previous keyboard shortcut (⌘-K) which we never figured out how to do before we eventually removed the feature. cc @ben

src/App.tsx
```diff
@@ -7,18 +7,111 @@ import './App.css'
...
+    ]
+  )
+
+  /* istanbul ignore next - need to figure out how to test this */
+  const onClick = useCallback(
+    ({ target }: React.MouseEvent) => {
+      if (target) {
+        const currentPath = window.location.pathname
+
+        setImmediate(() => {
```
beausmith: Is it safe to use `setImmediate`? 

- https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate
- https://caniuse.com/#feat=setimmediate

Perhaps we need to add `setImmediate` to our `package.json` dependencies?

src/App.tsx
```diff
@@ -7,18 +7,111 @@ import './App.css'
...
+
+  /* istanbul ignore next - need to figure out how to test this */
+  const onFocus = useCallback(
+    ({ target }: React.FocusEvent) => {
+      if (target) {
+        const currentPath = window.location.pathname
+
+        setImmediate(() => {
+          // Only send `onFocus` to the screen reader if the focus didn't
+          // trigger navigation.
```
beausmith: Where does `onFocus` trigger navigation in this codebase? Or is this just for safety?

src/App.tsx
```diff
@@ -7,18 +7,111 @@ import './App.css'
...
+    [screenReader]
+  )
+
+  /* istanbul ignore next - need to figure out how to test this */
+  const onFocus = useCallback(
+    ({ target }: React.FocusEvent) => {
+      if (target) {
+        const currentPath = window.location.pathname
+
+        setImmediate(() => {
```
beausmith: When referencing window methods we've been explicitly using the window method (`window.setTimeout(...)`) in order to clearly identify window methods from local methods.

@eventualbuddha - What do you think about this? Perhaps there is an ESLint rule we can add to enforce this?

src/components/LinkButton.tsx
beausmith: Super! There are at least 7 more places we use `window.setTimeout()` to avoid passing taps. Would love to figure out how to remove these as well… as they feel like hacks.

src/lib/gamepad.ts
beausmith: Yeah, this was related to legacy usage.

src/utils/ScreenReader.test.ts
beausmith: This looks like the function `h` is defined 3 times. How does this work?

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+   */
+  onPageLoad(): Promise<void>
+
+  /**
+   * Call this when the screen reader is enabled. Resolves when speaking is done.
+   */
+  onScreenReaderEnabled(): Promise<void>
+
+  /**
+   * Call this when the screen reader is disabled. Resolves when speaking is done.
```
beausmith: "Resolves when speaking is done." or immediately?

beausmith: Oh, I see. After speaking "Screen reader disabled" correct?

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+  private describeText(node: Text): string | undefined {
+    return node.textContent || undefined
+  }
+
+  private describeElement(node: Element): string | undefined {
+    if (this.isHidden(node)) {
+      return
+    }
+
+    const terminator = this.isBlockElement(node) ? '.' : ''
```
beausmith: I love this. I think we can safely remove many `aria-label` attributes where the value is the same as the text content plus a period.

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  private cleanDescription(
+    description: string | undefined
+  ): string | undefined {
+    if (!description) {
+      return
+    }
+    return description
+      .replace(/ {2}/g, ' ')
```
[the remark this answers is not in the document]
beausmith: Looks like you fixed this!

src/utils/ScreenReader.ts
```diff
@@ -0,0 +1,281 @@
...
+      .replace(/^[. ]+/g, '')
```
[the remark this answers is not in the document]
beausmith: I removed this in https://github.com/votingworks/bmd/pull/995

src/App.tsx
eventualbuddha: That’s interesting. It does work, so where is it coming from…?

src/utils/ScreenReader.ts
eventualbuddha: > Oh, I see. After speaking "Screen reader disabled" correct?

Yep.

src/utils/ScreenReader.test.ts
eventualbuddha: These are [function overloads](http://www.typescriptlang.org/docs/handbook/functions.html#overloads).

src/components/Modal.tsx
eventualbuddha: I could go either way. It’s not clear to me what the future of our ARIA markup should be. I’m okay removing it if you think we should.

src/App.tsx
eventualbuddha: Just for safety.

src/App.tsx
eventualbuddha: I don’t feel strongly about it. There already is an eslint rule that enforces things like `window.location` over `location`. Probably just need to expand that.

src/App.tsx
eventualbuddha: One way to do it is to move these out of `App` and into something a bit more testable, then maybe have a much smaller untested bit here.

.eslintrc.js
```diff
@@ -90,5 +90,6 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off', // Want to use it, but it requires return types for all built-in React lifecycle methods.
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-use-before-define': ['error', { functions: false }],
+    'import/extensions': 'off'
```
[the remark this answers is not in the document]
eventualbuddha: I was seeing errors with this in my local environment. Not sure why. It was complaining that we didn’t use `.ts` extensions.

src/components/Modal.tsx
beausmith: Ah right… we're not really using ARIA for anyone else other than ourselves. 

Well, I leave it to you to decide as perhaps you have the most recent experience understanding exactly what each of these does and how we benefit from them.

PR #994, opened 2019-12-19
fix: do not announce event targets that disappear

### LINKED ISSUE #985: Screen reader announces "Okay" button in modals briefly when dismissing
**Reproduction Steps**
1. Begin voting
2. Enable the screen reader by pressing `r`
3. Select a candidate in a 1-seat contest
4. Select another candidate in the same contest
5. You should see the modal telling you that you have to deselect the other one first
6. Press "Enter" to dismiss the modal

**Expected**
The current screen's default audio focus element should be read again.

**Actual**
The "Expected" does happen, but only after the "Okay" button is briefly announced as "O-" before being cut off when it starts announcing something else.

**Notes**
I had this same problem with navigation between contests with "Next" and "Back" and fixed it by triggering the screen reader only when an event didn't cause the location to change. We can't use that trick here because the location doesn't change, but maybe something similar would work.

_Originally posted by @eventualbuddha in https://github.com/votingworks/bmd/pull/966#issuecomment-567335172_

eventualbuddha:
When an event target, such as a modal dialog close button, is removed as a result of the event, we should not pass the event on to the screen reader.

Fixes #985

### DIFF
--- a/src/components/CandidateContest.tsx
+++ b/src/components/CandidateContest.tsx
@@ -275,25 +275,16 @@ class CandidateContest extends React.Component<Props, State> {
   }
 
   public closeAttemptedVoteAlert = () => {
-    // Delay to avoid passing tap to next screen
-    window.setTimeout(() => {
-      this.setState({ attemptedOvervoteCandidate: undefined })
-    }, 200)
+    this.setState({ attemptedOvervoteCandidate: undefined })
   }
 
   public confirmRemovePendingWriteInCandidate = () => {
-    // Delay to avoid passing tap to next screen
-    window.setTimeout(() => {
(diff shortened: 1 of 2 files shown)


src/components/CandidateContest.tsx
```diff
@@ -275,25 +275,16 @@ class CandidateContest extends React.Component<Props, State> {
   }
 
   public closeAttemptedVoteAlert = () => {
-    // Delay to avoid passing tap to next screen
-    window.setTimeout(() => {
-      this.setState({ attemptedOvervoteCandidate: undefined })
-    }, 200)
```
eventualbuddha: This required removing these delays similarly to what I did in #966. I couldn't reproduce the issue that these timeouts were trying to fix.

PR #995, opened 2019-12-19
more consistent audio instructions

beausmith:
While reviewing https://github.com/votingworks/bmd/pull/966 I noticed some inconsistencies with the audio instructions. This PR makes the instructions consistently state: "to do thing, use this button"

### DIFF
--- a/src/pages/StartPage.tsx
+++ b/src/pages/StartPage.tsx
@@ -43,7 +43,16 @@ const StartPage = () => {
             </h1>
             <hr />
             <p>
-              Your ballot has <strong>{contests.length} contests</strong>.
+              <span>
+                Your ballot has <strong>{contests.length} contests</strong>.
+              </span>
+              <span className="screen-reader-only">
+                When voting with the text-to-speech audio, use the accessible
+                controller to navigate your ballot. To navigate through the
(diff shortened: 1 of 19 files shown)


src/pages/StartPage.tsx
```diff
@@ -72,7 +81,7 @@ const StartPage = () => {
               primary
               to="/contests/0"
               id="next"
-              aria-label="Select next to start voting."
+              aria-label="Press the right button to advance to the first contest."
```
eventualbuddha: "right button" here vs. "down arrow" and "right arrow" in `ContestPage`. Which do we prefer?

beausmith: Thanks. Updated to use "button" everywhere.

beausmith:
Please always point out other inconsistencies. There's a lot of stuff to keep straight.

src/utils/ScreenReader.ts
```diff
@@ -169,7 +169,6 @@ export class AriaScreenReader implements ScreenReader {
     }
     return description
       .replace(/ +/g, ' ')
-      .replace(/^[. ]+/g, '')
```
eventualbuddha: What problem was this causing?

PR #1016, opened 2020-01-03
chore: forward BMD web port

eventualbuddha:
This makes it so you don't have to manually configure it in VS Code.


eventualbuddha:
> I wasn't doing this before (I think? Not on laptop now.) I want to know more about this.

This is only required if you're using the [remote development with SSH](https://code.visualstudio.com/docs/remote/ssh) setup in VS Code.

Issue #948, opened 2019-12-04
Update to TypeScript 3.7


eventualbuddha:
[The automated PR failed](https://github.com/votingworks/bmd/pull/947), but in a non-obvious way:

```
$ CI=true tsc && react-scripts test --maxWorkers 2 --env=jsdom --coverage && yarn test:e2e:ci
src/components/Loading.tsx:27:25 - error TS2589: Type instantiation is excessively deep and possibly infinite.

27       <ProgressEllipsis as={as} aria-label={`${children}.`}>
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is interesting, because [TS 3.7 adds support for recursive type aliases](https://github.com/microsoft/TypeScript/pull/33050). So why would upgrading to TS 3.7 cause what seems like an issue with recursive type aliases? I don't know, but I did some digging.

- https://github.com/jamesmfriedman/rmwc/issues/501 has a [reproduction](https://codesandbox.io/s/recursing-cdn-f00ic) of this problem, also involving `styled-components`.
- `@types/styled-components` was [updated for TS 3.7](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/b59a18caf3e954f6413f3cd8af85d2b1015dee29) and has version-specific types for `>= 3.7` that get rid of a workaround for the lack of non-recursive type aliases. It's possible that going back to the `< 3.7` types would fix the issue, but I haven't tested it.

PR #1021, opened 2020-01-03
chore(deps): upgrade react-scripts & TypeScript

eventualbuddha:
We can finally use nullish coalescing and optional chaining! I changed the obvious places in the codebase that could use them.

Closes #650, #948

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,7 +1,12 @@
+// Ensure we use all available babel parser plugins, but not the type annotation
+// ones since stylelint handles that itself.
+const { buildOptions } = require('@codemod/parser')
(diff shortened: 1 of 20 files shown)


.stylelintrc.js
```diff
@@ -1,7 +1,12 @@
+// Ensure we use all available babel parser plugins, but not the type annotation
+// ones since stylelint handles that itself.
+const { buildOptions } = require('@codemod/parser')
+const parserPlugins = buildOptions().plugins.filter(plugin => plugin !== 'typescript' && plugin !== 'flow')
```
eventualbuddha: This was needed because `nullishCoalescingOperator` isn't enabled by default in stylelint yet.

PR #1036, opened 2020-01-06
fix: don't allow double-tapping write-in "Accept"

### LINKED ISSUE #1028: Double-tap on "Add" button for write-ins causes all sorts of badness
There appears to be a delay when tapping "Accept" button on the write-in screen, during which the text of the write-in is cleared and the modal disappears after maybe 300-500ms.

If you tap "Accept" again in that delay (requires a quick double-tap on "Accept"), the result is two write-ins added and selected, one with the intended text, one that is empty. This happens even if it's a "vote for 1" question, meaning we break the validity of the answer. Once this happens, it's possible to keep selected candidates and adding more write-ins, as many as you want. Then, if one deletes one of the empty write-ins, it appears that the application crashes and goes blank (though this doesn't happen every time.)

eventualbuddha:
This was only reproducible with touch events (i.e. tapping) rather than click events. The fix is simply to remove the timeout that we previously needed but no longer do after the changes made to fix tap/click handling a while ago. We could also have added an explicit guard, but as this is now synchronous I didn't deem it worth the effort.

Fixes #1028

### DIFF
--- a/src/components/CandidateContest.test.tsx
+++ b/src/components/CandidateContest.test.tsx
@@ -129,7 +129,6 @@ describe('supports write-in candidates', () => {
     getByText('Write-In Candidate')
     typeKeysInVirtualKeyboard(getByText, 'LIZARD PEOPLE')
     fireEvent.click(getByText('Accept'))
-    jest.runOnlyPendingTimers() // Handle Delay when clicking "Accept"
     expect(queryByText('Write-In Candidate')).toBeFalsy()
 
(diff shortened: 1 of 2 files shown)


eventualbuddha:
> If you tested this and can confirm this no longer happens, then LGTM.

I tested it on a Chromebook, but I had to shrink the text down a lot to get the Accept button on top of a candidate. When I tapped the area where the buttons overlapped I did not get an unintended candidate selection.

PR #1042, opened 2020-01-08
refactor: consolidate `withMarkup` helpers

eventualbuddha:
This adapts the existing `withMarkup` to be flexible about the type returned by its query, allowing single or multiple elements to be found in a type-safe way.

### DIFF
--- a/test/helpers/withMarkup.ts
+++ b/test/helpers/withMarkup.ts
@@ -2,15 +2,16 @@
 
 import { MatcherFunction } from '@testing-library/react'
 
-type Query = (f: MatcherFunction) => HTMLElement
+type Query<T> = (f: MatcherFunction) => T
 
-const withMarkup = (query: Query) => (text: string): HTMLElement =>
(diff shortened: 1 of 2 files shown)


test/helpers/withMarkup.ts
```diff
@@ -2,15 +2,16 @@
 
 import { MatcherFunction } from '@testing-library/react'
 
-type Query = (f: MatcherFunction) => HTMLElement
+type Query<T> = (f: MatcherFunction) => T
 
-const withMarkup = (query: Query) => (text: string): HTMLElement =>
+const hasText = (text: string, node: HTMLElement) => node.textContent === text
```
eventualbuddha: Pulling this out wasn't strictly necessary, but I encountered this when trying to update `eslint` (which I remembered was the thing that made upgrading react-scripts so painful when I tried last time). The new version enforces [`no-shadow`](https://eslint.org/docs/rules/no-shadow), which we abuse a fair amount.

PR #1046, opened 2020-01-08
fix(keyboard): add aria labels to punctuation keys

eventualbuddha:
This ensures that the screen reader will read a label for punctuation keys such as "period" or "comma". Without this the screen reader simply says nothing.

### DIFF
--- a/src/components/VirtualKeyboard.tsx
+++ b/src/components/VirtualKeyboard.tsx
@@ -47,27 +47,79 @@ const Keyboard = styled.div`
 interface Props {
   onKeyPress: PointerEventHandler
   keyDisabled?(key: string): boolean
+  keyMap?: KeyMap
 }
 
-const VirtualKeyboard = ({ onKeyPress, keyDisabled }: Props) => (
+interface Key {
+  label: string
+  ariaLabel?: string
+}
+
+interface KeyMap {
+  rows: Key[][]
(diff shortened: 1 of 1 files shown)


src/components/VirtualKeyboard.tsx
```diff
@@ -47,27 +47,79 @@ const Keyboard = styled.div`
...
+      { label: 'C' },
+      { label: 'V' },
+      { label: 'B' },
+      { label: 'N' },
+      { label: 'M' },
+      { label: ',', ariaLabel: 'comma' },
+      { label: '.', ariaLabel: 'period' },
+      { label: '-', ariaLabel: 'dash' },
+    ],
+    [{ label: 'space' }, { label: '⌫ delete' }],
```
eventualbuddha: `⌫ delete` reads as "erase to the left, delete", which is a bit verbose but is okay IMO.

beausmith: Thanks for clarifying. Perhaps commit this a comment with this so that someone doesn't add and `ariaLabel` ?

eventualbuddha: I decided to add an `ariaLabel` with just `delete` since that's how it reads it on Linux now. On macOS it reads as "erase to the left, delete".

PR #1050, opened 2020-01-09
Setup error screens

### LINKED ISSUE #1041: Hardware Setup Error Screens and Warning Messages
This will support messaging for:
- low-power + not plugged in.
- card reader not plugged in.

beausmith:
Fixes #1041 

This PR needs:
- [x] backend data to enable screens.
- [x] tests based upon mocked data.

# Error Screens

These screens will display for all users if the state is such.

## Low battery and no charger
image

## Card Reader not detected
image

## No Printer Detected
image

## Printer Error Detected
image

# Warnings on Insert Card Screen

These warnings are informational but non-blocking. They are visible to both Poll Workers and Voters.

Note: It is possible for both of these warnings to appear at the same time.

## No Power Warning
image

## No Accessible Controller Warning
image

### DIFF
--- a/src/AppRoot.tsx
+++ b/src/AppRoot.tsx
@@ -52,6 +52,9 @@ import InvalidCardScreen from './pages/InvalidCardScreen'
 import InsertCardScreen from './pages/InsertCardScreen'
 import PollWorkerScreen from './pages/PollWorkerScreen'
 import PrintOnlyScreen from './pages/PrintOnlyScreen'
+import SetupCardReaderPage from './pages/SetupCardReaderPage'
+import SetupPrinterPage from './pages/SetupPrinterPage'
+import SetupPowerPage from './pages/SetupPowerPage'
 import UnconfiguredScreen from './pages/UnconfiguredScreen'
 import UsedCardScreen from './pages/UsedCardScreen'
 import { getBallotStyle, getContests, getZeroTally } from './utils/election'
@@ -61,6 +64,7 @@ import utcTimestamp from './utils/utcTimestamp'
 import { Card } from './utils/Card'
 import { Poller, IntervalPoller } from './utils/polling'
 import { Storage } from './utils/Storage'
+import { Hardware } from './utils/Hardware'
 
 interface CardState {
   isClerkCardPresent: boolean
@@ -88,6 +92,11 @@ interface SharedState {
   appPrecinctId: string
   ballotsPrintedCount: number
   election: OptionalElection
+  hasAccessibleControllerAttached: boolean
+  hasCardReaderAttached: boolean
+  hasChargerAttached: boolean
+  hasLowBattery: boolean
+  hasPrinterAttached: boolean
   isFetchingElection: boolean
   isLiveMode: boolean
   isPollsOpen: boolean
@@ -106,6 +115,7 @@ export interface AppStorage {
 }
 
 export interface Props extends RouteComponentProps {
+  hardware: Hardware
   card: Card
   storage: Storage<AppStorage>
   printer: Printer
@@ -120,6 +130,7 @@ class AppRoot extends React.Component<Props, State> {
   private machineIdAbortController = new AbortController()
 
   private cardPoller?: Poller
+  private statusPoller?: Poller
   private lastVoteUpdateAt = 0
   private lastVoteSaveToCardAt = 0
   private cardWriteInterval = 0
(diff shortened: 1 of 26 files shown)


src/AppRoot.tsx
```diff
@@ -793,11 +841,30 @@ class AppRoot extends React.Component<Props, State> {
...
+      hasChargerAttached,
+      hasLowBattery,
+      hasPrinterAttached,
+      hasPrinterError,
       precinctId,
       tally,
       userSettings,
       votes,
     } = this.state
+    if (hasLowBattery && !hasChargerAttached) {
```
eventualbuddha: All these will stop a voter who is in the middle of voting, right? Is that the behavior we want?

src/AppRoot.tsx
```diff
@@ -456,6 +471,37 @@ class AppRoot extends React.Component<Props, State> {
...
+          } catch (error) {
+            this.stopStatusPolling() // Assume backend is unavailable.
+          }
+        }
+      )
+    }
+  }
+
+  public stopStatusPolling = () => {
+    this.statusPoller && this.statusPoller.stop()
```
eventualbuddha:
```suggestion
    this.statusPoller?.stop()
```

src/AppRoot.tsx
beausmith: Yes. For loss of any functionality which will not allow a voter to complete their vote, we display one of these error screens. 

In the case of this specific error, if the charger is not plugged in and `!hasLowBattery`, the voter will see a non-blocking notification warning (TBD) on the Insert Card screen which is informational is design and has copy like: "This machine is on battery power. Please ask a poll worker to plug-in the power cord for this machine."

src/AppRoot.tsx
beausmith: Oh yeah! This was copy/pasta/edit, but should have caught it.

beausmith:
@eventualbuddha - I have added a Hardware API. This seemed simpler than separate APIs for each. 

Open to all suggestions for any refactors, etc.

What else do you need from me to implement a "module-hardware" or whatever you think is best?

src/utils/Hardware.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /**
+   * Reads Accessible Controller status
+   */
+  public async readAccesssibleControllerStatus(): Promise<
+    AcessibleControllerAPI
+  > {
+    console.log("Why is 'this' undefined?", this)
+    return this.accessibleControllerStatus
```
beausmith: @eventualbuddha - The last commit on this branch is a WIP with debugging code. I'm stumped as to why `this` is undefined. `MemoryHardware` is modeled after the `MemoryCard`, but something must be different. Do you see where the error in this code is?

src/AppRoot.tsx
```diff
@@ -456,6 +472,55 @@ class AppRoot extends React.Component<Props, State> {
...
+        async () => {
+          try {
+            // Possible implementation
+            const {
+              readAccesssibleControllerStatus,
+              readBatteryStatus,
+              readCardReaderStatus,
+              readPrinterStatus,
+            } = this.props.hardware
+            const accesssibleController = await readAccesssibleControllerStatus()
```
eventualbuddha: This is the call that’s making `this` be `undefined` inside `readAccesssibleControllerStatus`. When you call a function without a context in strict mode you get `this === undefined` (in non-strict mode you get `this === window`).

For example:
```ts
const obj = { m() { return this } }
const { m } = obj
const mBound = ::obj.m // https://github.com/tc39/proposal-bind-operator

// These all pass.
assert(obj.m() === obj)
assert(m() === undefined)
assert(m.call(obj) === obj)
assert(mBound() === obj)
```

Note that the bind operator proposal is still stage 0 and [won’t be implemented in TS until stage 3 at the earliest](https://github.com/Microsoft/TypeScript/issues/3508).

src/utils/Hardware.ts
```diff
@@ -0,0 +1,246 @@
...
+
+  /**
+   * Sets Battery charging
+   */
+  setBatteryCharging(charging: boolean): Promise<void>
+
+  /**
+   * Sets Battery level.  Number between 0–1.
+   */
+  setBatteryLevel(level: number): Promise<void>
```
eventualbuddha: These `set*` methods don’t represent operations we can actually perform in non-test scenarios, i.e. the battery level is what it is. I’d say these should not be present on the `Hardware` interface but should be defined on `MemoryHardware`. Unless I’m missing something?

src/utils/Hardware.ts
eventualbuddha: Commented at the source of the problem. It’s an unbound method problem. I’d fix it by doing `const { hardware } = this.props` and calling i.e. `await hardware.read…`.

beausmith:
@eventualbuddha - I've rebased, squashed, and pushed commits. Everything is working, however there are a couple regressions I've failed to figure out. I've got to hang my hat for the night. Back to this tomorrow.

PR #1053, opened 2020-01-09
fix: prefer US English voices in speech synthesis

eventualbuddha:
This ensures we don't end up with an unusual default, such as "Afrikaans espeak-ng".


eventualbuddha:
The voices vary by OS and by browser. In production we really only care about the specific CMU English voice that we install on VxMark. For development and demo, this really just tries to find a sane US English voice. For example, on macOS Safari, Firefox, and Chrome all use the default "Alex" voice. Chrome would actually have used the "Google US English" voice except that it's not a local voice.

Really all this does is do a best guess but will always pick _something_ if there's any local voices to pick from.

PR #1054, opened 2020-01-09
fix: ensure unique keyboard row `key` attribute

eventualbuddha:
Adding a level of nesting to the data structure didn't affect the `join` since it still runs and type checks, but it stopped doing the right thing. Namely, the keys looked like `row-[object Object],[object Object]…`. Now they again look like `row-Q,W,E…`.


eventualbuddha:
> Sorry I missed this!

It was caused by [my commit](https://github.com/votingworks/bmd/commit/a00ca2ca4cb7079202cda9e5b68ae3706019f9cd), and you'd have to have looked at the test/dev logs to see it.

PR #1079, opened 2020-01-15
feat: add support for kiosk-browser

eventualbuddha:
This adds types and test mocking support for kiosk-browser.

### DIFF
--- a/src/types/kiosk-browser.d.ts
+++ b/src/types/kiosk-browser.d.ts
@@ -1,0 +1,20 @@
+declare module 'kiosk-browser' {
+  export interface BatteryInfo {
+    discharging: boolean
+    level: number // Number between 0–1
+  }
+
+  export interface Kiosk {
+    print(): Promise<void>
+    getBatteryInfo(): Promise<BatteryInfo>
+  }
+}
+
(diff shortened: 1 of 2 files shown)


src/types/kiosk-browser.d.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// Disable `no-var` because using `var` ensures `kiosk` is a property on
+// `globalThis`, which makes it available both as plain `kiosk` and as
+// `window.kiosk`. Using `const` or `let`, as eslint suggests, will not make it
+// available at all. An alternative would be to add `kiosk` as a property to
+// the `Window` interface, but then we couldn't refer to `kiosk` without
+// `window`.
+// eslint-disable-next-line no-var
+declare var kiosk: import('kiosk-browser').Kiosk | undefined
```
beausmith: FWIW, I'm a fan of us using `window.kiosk` as it adds more clarity when reading code etc. But I could be convinced  otherwise.

eventualbuddha: I'm fine with that. In reality there'll likely be only a handful of places we use it.

Issue #978, opened 2019-12-17
Initialize empty ballot on voter card when "Start Voting" button is pressed


benadida:
Repro: skip every question on ballot --> empty ballot on card

beausmith:
**Suggested Solution:** Clicking "Start Voting" button initializes an empty ballot value on the card. Ballot initialization happen upon card insert as it's not clear that the voter intended to do anything.

PR #1088, opened 2020-01-17
force empty-vote write to the card on "start voting" button press

benadida:
solves #978

### DIFF
--- a/src/AppEndToEnd.test.tsx
+++ b/src/AppEndToEnd.test.tsx
@@ -171,6 +171,10 @@ it('VxMark+Print end-to-end flow', async () => {
   // Start Voting
   fireEvent.click(getByText('Start Voting'))
 
+  // Initial empty votes written to the card after tapping "Start Voting".
+  await advanceTimersAndPromises()
+  expect(writeLongUint8ArrayMock).toHaveBeenCalledTimes(1)
+
   // Advance through every contest
   for (let i = 0; i < voterContests.length; i++) {
     const title = voterContests[i].title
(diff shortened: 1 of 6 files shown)


benadida:
not quite ready to merge as it seems to be causing two writes in a row on initialization. Write / Read / Write long value. Want to prevent that before committing.

eventualbuddha:
@benadida I rebased this and got the [error you can see in CI](https://app.circleci.com/jobs/github/votingworks/bmd/2385):

```
 FAIL  src/AppEndToEnd.test.tsx (7.811s)
  ● VxMark+Print end-to-end flow

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 1
    Received number of calls: 2

      189 |       advanceBy(1100)
      190 |       await advanceTimersAndPromises()
    > 191 |       expect(writeLongUint8ArrayMock).toHaveBeenCalledTimes(1)
          |                                       ^
      192 | 
      193 |       // If we wait another second and advance timers, without any change made to the card,
      194 |       // we should not see another call to save the card data

      at Object.toHaveBeenCalledTimes (src/AppEndToEnd.test.tsx:191:39)
```

Looks like the changes I made to the end-to-end tests now correctly wait for the new card write you're doing. To fix this I just added an assertion about empty votes being written and then incremented the expected call count.

benadida:
> Looks like the changes I made to the end-to-end tests now correctly wait for the new card write you're doing. To fix this I just added an assertion about empty votes being written and then incremented the expected call count.

thank you!

PR #1105, opened 2020-01-22
feat: enable speech by default from speech-loader entry point

eventualbuddha:
When speech-loader.html is able to load voices properly, it will load index.html with a query string parameter telling it to enable the screen reader right away.

The API needed to change to support this. Instead of swapping out `AriaScreenReader` instances with different text-to-speech implementations (null or real), we add the ability to mute TTS instances. This actually maps pretty well to the current volume control in the speech synthesis API which we may want to use in the future anyway. So now we keep the screen reader instance and just enable/disable it.

In addition, a number of tests were not waiting properly for the async methods they were calling. The tests worked anyway, but this fixes it to make the more correct.

### DIFF
--- a/public/speech-loader.html
+++ b/public/speech-loader.html
@@ -16,11 +16,13 @@
 
       if (timeout <= now) {
         console.log('timeout exceeded waiting for voices, continuing anyway')
-        setTimeout(goToApp, oneSecond)
+        setTimeout(() => goToApp('reader=off'), oneSecond)
         return true
       } else {
         const seconds = Math.floor((timeout - now) / oneSecond)
-        console.log(`timeout not yet reached, still waiting for voices for ${seconds}s`)
+        console.log(
+          `timeout not yet reached, still waiting for voices for ${seconds}s`
+        )
       }
     } else {
       const timeout = Date.now() + secondsToWaitForVoices * oneSecond
@@ -32,8 +34,8 @@
     return false
   }
 
-  function goToApp() {
-    location.href = '/'
+  function goToApp(query = undefined) {
+    location.href = query ? `/?${query}` : '/'
   }
 
   /**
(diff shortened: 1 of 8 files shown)


public/speech-loader.html
```diff
@@ -16,11 +16,13 @@
 
       if (timeout <= now) {
         console.log('timeout exceeded waiting for voices, continuing anyway')
-        setTimeout(goToApp, oneSecond)
+        setTimeout(() => goToApp('reader=off'), oneSecond)
         return true
       } else {
         const seconds = Math.floor((timeout - now) / oneSecond)
-        console.log(`timeout not yet reached, still waiting for voices for ${seconds}s`)
+        console.log(
```
benadida: I'm okay leaving these console messages in, just want to make sure you meant to.

src/App.tsx
```diff
@@ -30,51 +28,26 @@ export interface Props {
...
-  }
+  screenReader?: ScreenReader
 }
 
 const App = ({
-  tts = {
-    enabled: new SpeechSynthesisTextToSpeech(memoize(getUSEnglishVoice)),
-    disabled: new NullTextToSpeech(),
-  },
+  screenReader = new AriaScreenReader(
```
benadida: nice, way cleaner.

src/index.tsx
```diff
@@ -5,9 +5,30 @@ import App from './App'
...
 import * as serviceWorker from './serviceWorker'
+import {
+  AriaScreenReader,
+  SpeechSynthesisTextToSpeech,
+} from './utils/ScreenReader'
+import memoize from './utils/memoize'
+import { getUSEnglishVoice } from './utils/voices'
+
+const isSampleApp = window.location.hash === '#sample'
+const readerEnabled = window.location.search === '?reader=on'
```
benadida: this is gonna bust if we ever add another parameter. How about using:
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

benadida: I'm also wondering what happens if we ever reload the page for any reason (a fault, etc.) Don't have to solve it now, but maybe flag that for later.

eventualbuddha: Good idea, done.

public/speech-loader.html
eventualbuddha: Yes, figured they'd be useful on this screen.

PR #1113, opened 2020-01-24
fix: correct some issues with polling

eventualbuddha:
Firstly, `setInterval` was causing calls to the polling callback to stack on top of each other. WIth polling hardware, this meant hammering the CPU. Now it waits until the callback resolves when a promise is returned.

Second, I think this is what was causing issues with our tests timing out. I realized they were related because the end-to-end tests started failing when I fixed `IntervalPoller`. This is because they were not waiting for the poller to schedule the next callback in the `finally`, which was happening after a promise tick. So now wherever we had `advanceTimers` I made a new call to `await advanceTimersAndPromises` that optionally advances time and also awaits promises to ensure that anything already resolved had a chance to continue. These tests now run in a reasonable amount of time.

### DIFF
--- a/src/APICardCatch.test.tsx
+++ b/src/APICardCatch.test.tsx
@@ -1,9 +1,9 @@
 import React from 'react'
-import { fireEvent, render, wait } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 
 import App from './App'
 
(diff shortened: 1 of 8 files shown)


beausmith:
Wow. This makes so much sense. Thanks @eventualbuddah

PR #1116, opened 2020-01-27
Updated Printed Ballot design and Electionguard tracking code page

beausmith:
I've separated these commits such that the two related to "Printed Ballot design" can be easily cherry-picked to master. 

The commits related to "Electionguard tracking code page" do not have updated tests as the base `electionguard` branch does nor either..


src/endToEnd.ts
```diff
@@ -3,7 +3,7 @@ import { VotesDict } from '@votingworks/ballot-encoder'
 import { E2EAPI } from './config/types'
 import fetchJSON from './utils/fetchJSON'
 
-export default async function encryptAndGetTracker(
+export default async function getBallotTrackingCode(
```
benadida: this new name hides the fact that the ballot is being encrypted and recorded by the EG backend. Specifically, you don't want to call this function twice on the same ballot. So I would suggest a different name than something that sounds like a getter.

src/pages/PrintOnlyScreen.tsx
```diff
@@ -143,7 +142,7 @@ const PrintOnlyScreen = ({
             />
           </p>
           <h1>
-            <Loading>Printing your official ballot & tracker</Loading>
+            <Loading>Printing your official ballot</Loading>
```
benadida: I'd like to keep the indication that the tracker is being printed. I can see this being split into a second screen at some later point, but for now keeping the indication in this text seems best to me.

src/components/ElectionGuardBallotTrackingCode.tsx
```diff
@@ -0,0 +1,140 @@
...
+const ElectionGuardBallotTrackingCode = ({
+  election,
+  tracker,
+}: {
+  election: Election
+  tracker: string
+}) => {
+  const { title, county, state, date } = election
+
+  const trackingDisplayUrl = 'https://fulton.electionguard.vote'
```
danwallach: Do you want to hard-code this URL?

src/components/ElectionGuardBallotTrackingCode.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const trackerSafeForEmbedding = tracker.replace(/ /g, '-')
+  const trackerUrl = trackingTemplateUrl.replace(
+    '<tracker_id>',
+    trackerSafeForEmbedding
+  )
+
+  return (
+    <TrackerPage>
+      <PreHeader>
+        This is your ballot tracking code. Take this with you.
```
danwallach: Do we want to hard-code English strings like this, versus having some sort of multi-language strategy?

src/pages/PrintOnlyScreen.tsx
```diff
@@ -50,7 +50,7 @@ const PrintOnlyScreen = ({
   const printerTimer = useRef(0)
   const [okToPrint, setOkToPrint] = useState(true)
   const [isPrinted, updateIsPrinted] = useState(false)
-  const [trackerString, setTrackerString] = useState<string | undefined>('')
+  const [trackerString, setTrackerString] = useState('')
```
danwallach: So I'm trying to wrap my head around this. The initial value of the tracking string is the empty-string, which gets set by printBallot, which is inside PrintOnlyScreen. I'm just trying to convince myself that the tracker is always set before the printing happens. That seems to bottom out to a call to the ElectionGuard server (/electionguard/EncryptBallot), which may or may not be idempotent (i.e., it might return something different each time and/or have other side-effects).

On the one hand, I'd say a little bit of paranoia is appropriate here, in the rare case that printing is triggered twice. Maybe you want to hang onto the result from the ElectionGuard server, but then you have to also make sure you throw it away once the user session is done.

(Or, maybe I'm just over-thinking things?)

src/components/ElectionGuardBallotTrackingCode.tsx
beausmith: Yes, we want to do i18n + l10n in a future milestone. One "feature" of our initial market is that they only require a single language for ballots.

src/components/ElectionGuardBallotTrackingCode.tsx
beausmith: @ben, thoughts on moving this to election config JSON?

src/endToEnd.ts
beausmith: Hmm… I was trying to hide this. The front end only needs a ballot tracking code and doesn't need to know about what is happening under the hood, no?

src/pages/PrintOnlyScreen.tsx
```diff
@@ -62,9 +62,8 @@ const PrintOnlyScreen = ({
     !isPrinted
 
   const printBallot = useCallback(async () => {
-    // get the tracker first
-    const tracker = await encryptAndGetTracker(votes)
-    setTrackerString(tracker)
+    const ballotTrackingCode = await getBallotTrackingCode(votes)
+    setTrackerString(ballotTrackingCode)
```
eventualbuddha: @danwallach flagged an interesting issue above that made me wonder about `trackerString`. I'm guessing that if `markVoterCardPrinted` were synchronous that `trackerString` would always be empty when printing. As-is, I think it constitutes a race condition since React rendering after mutating state is also asynchronous.

benadida: Ahah, good catch! What would be the right way to wait until the re-rendering?

src/components/QRCode.test.tsx
```diff
@@ -3,14 +3,22 @@ import { render } from '@testing-library/react'
...
-    />
-  `)
+describe('renders QR Code', () => {
+  it('renders basic QR Code', async () => {
+    const { container } = render(
+      <QRCode size="1in" value="http://voting.works" />
+    )
+    expect(container.firstChild).toMatchSnapshot()
+  })
+  it('renders QR Code with size', async () => {
```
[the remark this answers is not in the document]
beausmith: 🤦🏼‍♂️

src/pages/PrintOnlyScreen.tsx
eventualbuddha: I decided to model this as a state machine like so:

```
Wait for Card
- card inserted with votes -> Encrypt Ballot
- card inserted without votes -> Empty Card

Encrypt Ballot
- ballot encrypt succeeded -> Mark Card Used
- ballot encrypt failed -> ???

Mark Card Used
- erase card succeeded -> Print
- erase card failed -> ???

Print
- print started -> Update Tally
- print failed to start -> ???

Update Tally
- update tally succeeded -> Printing
- update tally failed -> ???

Printing
- print succeeded -> Verify Printed Ballot
- print failed -> ???

Verify Printed Ballot
- card removed -> Wait for Card

Empty Card
- card removed -> Wait for Card
```

Each state is named e.g. `Empty Card` and represents a particular React render. The actions that can happen in that state are enumerated below the name along with the state they should transition to. Notice that there are several error handling states that are missing.

> Ahah, good catch! What would be the right way to wait until the re-rendering?

I think we need to set the tracker code state and then, in another render, trigger the print. While I don't think we _have_ to make the states more explicit in this PR, I do think it might help.

src/components/ElectionGuardBallotTrackingCode.tsx
benadida: eventually yes, but for WI, I want to limit the number of repos we change.

beausmith:
@benadida - I've made the updates to this branch which you requested verbally. Please feel free to review and merge if you're satisfied.

Issue #1087, opened 2020-01-16
Create provider to return `machine-id` and `appMode` values


beausmith:
Maybe `getMachineInfo()` returns `{ id: 7, mode: 'mark+print' }`

Or…`{ id: 7, mark: true, print: false }`

`AppSetupErrors.test.tsx` returns three of the following errors which are calls to `/machine-id`.
```
console.error node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/virtual-console.js:29
      Error: Error: connect ECONNREFUSED 127.0.0.1:80
```

PR #1121, opened 2020-01-27
refactor: inject `machineId` provider

### LINKED ISSUE #714: Update fetchMock.restore() strategy
Because of the following in `setupTests.tsx`…

```js
afterEach(() => {
  fetchMock.restore()
})
```

…the following had to be added to `AppMarkVoterCardInvalid.test.tsx`:

```js
    // TODO: This is required due to `fetchMock.restore()` in `setupTests.tsx`.
    fetchMock.get('/machine-id', () => JSON.stringify({ machineId: '1' }))
```

eventualbuddha:
Rather than making `AppRoot` fetch `/machine-id` and then using `fetchMock` to mock it everywhere, we inject a provider for `machineId` that is more easily mocked.

This also cleans up our use of `fetchMock` by ensuring that we mock `fetch` before _every_ test and restore behavior after _every_ test. This way we won't accidentally make real requests in tests that show up in error logs as e.g. `connect ECONNREFUSED 127.0.0.1:80`.

Refs #1087, does not do `appMode` yet
Closes #714


eventualbuddha:
> Do you think we'll refactor `machineId` to `machine.id` when we migrate `appMode` into this, potentially as `machine.mode`?

Yes.

PR #1124, opened 2020-01-27
Monitor accessible controller connection status

eventualbuddha:
This should reduce CPU overhead a bit, and is generally the approach I'd like to move everything to that currently polls.

### DIFF
--- a/src/AppRoot.tsx
+++ b/src/AppRoot.tsx
@@ -16,6 +16,8 @@ import 'normalize.css'
 import React from 'react'
 import Gamepad from 'react-gamepad'
 import { RouteComponentProps } from 'react-router-dom'
+// eslint-disable-next-line import/no-unresolved
+import { Listener, ChangeType, Device } from 'kiosk-browser'
 import './App.css'
 import Ballot from './components/Ballot'
 import * as GLOBALS from './config/globals'
@@ -63,7 +65,7 @@ import utcTimestamp from './utils/utcTimestamp'
 import { Card } from './utils/Card'
 import { Poller, IntervalPoller } from './utils/polling'
 import { Storage } from './utils/Storage'
-import { Hardware } from './utils/Hardware'
+import { Hardware, isAccessibleController } from './utils/Hardware'
 
 interface CardState {
   isClerkCardPresent: boolean
@@ -131,6 +133,7 @@ class AppRoot extends React.Component<Props, State> {
 
   private cardPoller?: Poller
   private statusPoller?: Poller
+  private onDeviceChangeListener?: Listener<[ChangeType, Device]>
   private lastVoteUpdateAt = 0
   private lastVoteSaveToCardAt = 0
   private forceSaveVoteFlag = false
@@ -163,7 +166,7 @@ class AppRoot extends React.Component<Props, State> {
     appPrecinctId: '',
     ballotsPrintedCount: 0,
     election: undefined,
-    hasAccessibleControllerAttached: true,
+    hasAccessibleControllerAttached: false,
     hasCardReaderAttached: true,
     hasChargerAttached: true,
     hasLowBattery: false,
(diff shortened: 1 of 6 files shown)


src/AppRoot.tsx
```diff
@@ -497,11 +497,40 @@ class AppRoot extends React.Component<Props, State> {
...
+      ;(async () => {
+        // get initial status
+        const accessibleControllerStatus = await hardware.readAccesssibleControllerStatus()
+
+        this.setState({
+          hasAccessibleControllerAttached: accessibleControllerStatus.connected,
+        })
+
+        // watch for changes
+        this.onDeviceChangeListener = hardware.onDeviceChange.add(
```
eventualbuddha: This won’t be set by the time the method exits, which means this could be added twice. I should address that.

src/AppRoot.tsx
```diff
@@ -497,11 +497,40 @@ class AppRoot extends React.Component<Props, State> {
...
+          (changeType, device) => {
+            /* istanbul ignore else */
+            if (isAccessibleController(device)) {
+              this.setState({
+                hasAccessibleControllerAttached:
+                  changeType === 0 /* ChangeType.Add */,
```
eventualbuddha: I can’t use the actual enum value here because it doesn’t exist as a value, only a type. It’s defined in a TypeScript declaration file. I should fix that, but maybe not in this commit.

src/utils/Hardware.ts
```diff
@@ -126,24 +159,105 @@ export class MemoryHardware implements Hardware {
...
- */
-export class KioskHardware extends MemoryHardware {
-  public constructor(private kiosk: Kiosk) {
-    super()
-    this.kiosk = kiosk
+  /**
+   * Manage notifications for device changes.
+   */
+  public onDeviceChange: Listeners<[ChangeType, Device]> = (() => {
+    const listeners = new Set<
```
eventualbuddha: Call this `callbacks`?

src/AppRoot.tsx
eventualbuddha: Yeah, I'm going to leave this for later.

eventualbuddha:
This requires https://github.com/votingworks/kiosk-browser/pull/14.

src/AppRoot.tsx
```diff
@@ -513,11 +513,36 @@ class AppRoot extends React.Component<Props, State> {
         }
       )
     }
+
+    /* istanbul ignore else */
+    if (!this.onDeviceChangeListener) {
+      const { hardware } = this.props
+
+      this.setState({
+        hasAccessibleControllerAttached: false,
```
eventualbuddha: This should probably just be the default value.

eventualbuddha:
> Wondering why using direct integers rather than `ChangeType.Add`?

This is because the `ChangeType` enum is defined in a `.d.ts` file and therefore has no runtime/value part, only a typing part. Long term I'd like to consolidate these types into a package that both `bmd` and `kiosk-browser` depend on.

benadida:
> > Wondering why using direct integers rather than `ChangeType.Add`?
> 
> This is because the `ChangeType` enum is defined in a `.d.ts` file and therefore has no runtime/value part, only a typing part. Long term I'd like to consolidate these types into a package that both `bmd` and `kiosk-browser` depend on.

Got it. Not a blocker at all, but IMO the benefit of typing this is less than the cost of code clarity. But I can live with it.

eventualbuddha:
> IMO the benefit of typing this is less than the cost of code clarity. But I can live with it.

As in, it'd be clearer to just have `0` and `1`?

benadida:
I mean that defining `const` instead of types and having code say `=== MyConst` feels clearer. Again, this is a nit, I'm fine with this :)

eventualbuddha:
Ah, okay. We can do both.

PR #1125, opened 2020-01-28
tweaks to make e2e work against current build of the EG backend

benadida:
The current EG backend expects calls to `/election/*` and requires the right content-type.

These changes are confirmed to work against the latest EG backend.

@beausmith this can be merged now or after your changes go in, I don't have any problems rebasing.

### DIFF
--- a/src/endToEnd.ts
+++ b/src/endToEnd.ts
@@ -1,20 +1,24 @@
-import { VotesDict } from '@votingworks/ballot-encoder'
+import { CompletedBallot } from '@votingworks/ballot-encoder'
 
 import { E2EAPI } from './config/types'
 import fetchJSON from './utils/fetchJSON'
 
-export default async function encryptBallot(
-  // eslint-disable-next-line
-  votes: VotesDict
-) {
+export default async function encryptBallot(ballot: CompletedBallot) {
   try {
     const { tracker } = await fetchJSON<E2EAPI>(
       '/electionguard/EncryptBallot',
       {
         method: 'post',
-        body: JSON.stringify(votes),
+        body: JSON.stringify({
+          ballot: ballot,
+          currentBallotCount: 0,
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
       }
     )
+
     return tracker
   } catch (error) {
     return ''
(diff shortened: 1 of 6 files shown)


src/endToEnd.ts
```diff
@@ -12,7 +12,13 @@ export default async function encryptAndGetTracker(
       '/electionguard/EncryptBallot',
       {
         method: 'post',
-        body: JSON.stringify(votes),
+        body: JSON.stringify({
+          ballot: votes,
```
beausmith: image

beausmith: 😂 Sorry, couldn't help it.

benadida: lol nice :)

src/endToEnd.ts
```diff
@@ -1,20 +1,24 @@
...
+export default async function encryptBallot(ballot: CompletedBallot) {
   try {
     const { tracker } = await fetchJSON<E2EAPI>(
       '/electionguard/EncryptBallot',
       {
         method: 'post',
-        body: JSON.stringify(votes),
+        body: JSON.stringify({
+          ballot: ballot,
+          currentBallotCount: 0,
```
eventualbuddha: What is this?

src/pages/PrintOnlyScreen.tsx
```diff
@@ -62,7 +72,15 @@ const PrintOnlyScreen = ({
...
   const printBallot = useCallback(async () => {
-    const ballotTrackingCode = await encryptBallot(votes)
+    const ballotTrackingCode = await encryptBallot({
+      election,
+      ballotStyle: getBallotStyle({ ballotStyleId, election }),
+      precinct: getPrecinctById({ precinctId, election })!,
+      ballotId,
+      votes,
+      isTestBallot: !isLiveMode,
+      ballotType: 0,
```
eventualbuddha: This is an enum value, right? Can we use `BallotType.Normal` or whatever this is?

src/pages/PrintOnlyScreen.tsx
```diff
@@ -71,11 +89,22 @@ const PrintOnlyScreen = ({
     if (isUsed) {
       await printer.print()
       updateTally()
+      setBallotId('')
```
eventualbuddha: Won’t this just immediately generate a new one because of your conditional after the `useState` initializer?

src/endToEnd.ts
benadida: this is the total number of currently cast ballots, for sync'ing with EG backend. I'm setting it to 0 for now as I think it's unused, but will improve this.

src/pages/PrintOnlyScreen.tsx
benadida: no longer relevant given Beau's PR, but the idea here is yeah to say "done with this ballot Id" and have the earlier conditional regenerate a new one the next time it's rendered for the next ballot.

benadida:
replacing this PR with a much simpler one given @beausmith's work.

PR #1134, opened 2020-01-29
test: ensure calls to /card/reader don't leak

eventualbuddha:
We should not be using the "real" hardware classes in tests, so this uses `MemoryHardware` in the `AppData` tests. Also, adds a test to ensure `WebBrowserHardware#readCardReaderStatus` works as expected.


eventualbuddha:
> By "leak" do you mean leak an HTTP call?

Yes. `fetch-mock` complained about an unmatched HTTP call.

PR #1138, opened 2020-01-30
feat: monitor card reader via USB device changes

eventualbuddha:
This replaces the polling of `/card/reader` with a device add notification that checks for the specific card reader we use in production. The upside is that plugging or unpluging triggers updates immediately. The main downside is that this hardcodes a specific product rather than being generalized to any card reader that psycard supports. I think the trade-off is worth it, but I'm open to other opinions.

### DIFF
--- a/src/AppRoot.tsx
+++ b/src/AppRoot.tsx
@@ -64,7 +64,7 @@ import { getBallotStyle, getContests, getZeroTally } from './utils/election'
 import { Printer } from './utils/printer'
 import utcTimestamp from './utils/utcTimestamp'
 import { Card } from './utils/Card'
-import { Poller, IntervalPoller } from './utils/polling'
+import IntervalPoller from './utils/IntervalPoller'
 import { Storage } from './utils/Storage'
 import { Hardware, isAccessibleController } from './utils/Hardware'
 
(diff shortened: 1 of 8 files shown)


benadida:
I'd prefer that we discuss this before merging, for example I'm not entirely sure that every model of card reader we use has the exact same device ID. There are apparently different versions of our card reader that look the same.

eventualbuddha:
> I'm not entirely sure that every model of card reader we use has the exact same device ID.

Do you know whether that applies to other things like the accessible controller and printers?

benadida:
> > I'm not entirely sure that every model of card reader we use has the exact same device ID.
> 
> Do you know whether that applies to other things like the accessible controller and printers?

Great question... I believe those models are more stable, or at least I've seen no evidence that they aren't, whereas I've explicitly seen different model numbers for our card readers (and even different prices for them). I don't know if that means different USB identifiers, I just worry it might.

eventualbuddha:
I think I’ll redo this to be similar to what I just did for printing and still call `/card/reader` when I see a USB change. Then we won’t rely on ids at all.

benadida:
> I think I’ll redo this to be similar to what I just did for printing and still call `/card/reader` when I see a USB change. Then we won’t rely on ids at all.

I love this approach. In general we should monitor for all USB devices, and if we see one we don't like, we're going to take drastic action at some point. So "monitor USB devices" and then dispatch to `/card/reader` sounds perfect.

src/AppRoot.tsx
```diff
@@ -673,7 +675,19 @@ class AppRoot extends React.Component<Props, State> {
...
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    const {
+      hasAccessibleControllerAttached,
+      hasCardReaderAttached,
+      hasChargerAttached,
+      hasPrinterAttached,
+      hasLowBattery,
+      ...rest
+    } = this.initialState
+    /* eslint-enable @typescript-eslint/no-unused-vars */
```
eventualbuddha: We should replace this with something that is persistent and lives outside of the component tree.

eventualbuddha: Created #1152.

PR #1139, opened 2020-01-30
Invalid voter card data

### LINKED ISSUE #1055: voter card with invalid precinct / ballot style crashes BMD
let's not do that.

beausmith:
Fixes #1055 

Display new error screen when voter card has ballot style id or precinct id which doesn't match those in the current election definition.


beausmith:
@eventualbuddha - I have 'em. I will submit PR.

PR #1149, opened 2020-01-31
Separate printing of ballot and tracking code.

beausmith:
I took inspiration from @eventualbuddha's [comment](https://github.com/votingworks/bmd/pull/1116#discussion_r372519895) to implement with a state machine… but also took some shortcuts for time.

Also, pls note that I have purposely ignored the `PrintPage` used when device `appMode === "mark+print"`

separate-printing-of-ballot-and-tracking-code

### DIFF
--- a/src/components/PrintedBallot.tsx
+++ b/src/components/PrintedBallot.tsx
@@ -185,6 +185,7 @@ const YesNoContestResult = (props: {
   )
 
 interface Props {
+  ballotId?: string
   ballotStyleId: string
   election: Election
   isLiveMode: boolean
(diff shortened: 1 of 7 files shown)


src/components/PrintedBallot.tsx
```diff
@@ -193,13 +194,13 @@ interface Props {
 }
 
 const PrintBallot = ({
+  ballotId = randomBase64(),
```
benadida: I think we should require `ballotId`, otherwise there's a chance we code this incorrectly and generate new ballot IDs on the fly, which is never the right answer.

beausmith: Agreed. There are two other places where this is used. I didn't have the capacity to make these changes last night.

PR #1153, opened 2020-01-31
adapt EG calls to actual backend

benadida:
adjusted backend calls and data structure for EG to work with actual backend code.


eventualbuddha:
Something bad happened in a git rebase, I think.

benadida:
> Something bad happened in a git rebase, I think.

fixed :)

PR #1159, opened 2020-02-01
simplify and fix state machine for printing

benadida:
The previous implementation had an issue where the ballot id was being set and then, without a render cycle, the encryption was being done, which caused the ballot ID that was sent to the encryption to be the *old* ballot ID, starting with the empty string for the first one.

In trying to fix this, I realized that the state machine we need here is so simple that using the full-blown state machine package seems like significant overkill. So I tried the following simple-string-state-and-useEffect implementation, which seems to work like a charm.

### DIFF
--- a/src/pages/PrintOnlyScreen.tsx
+++ b/src/pages/PrintOnlyScreen.tsx
@@ -1,7 +1,5 @@
 import React, { useEffect, useRef, useState } from 'react'
 import styled from 'styled-components'
-import { useMachine } from '@xstate/react'
-import { Machine } from 'xstate'
 
 import { VotesDict, Election } from '@votingworks/ballot-encoder'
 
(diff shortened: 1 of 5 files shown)


beausmith:
@benadida - run and commit changes from `yarn uninstall @xstate/react xstate`

src/pages/PrintOnlyScreen.tsx
```diff
@@ -92,7 +54,7 @@ const PrintOnlyScreen = ({
   const printBallotTimer = useRef(0)
   const printBallotTrackingCodeTimer = useRef(0)
 
-  const [stateMachine, sendStateMachineEvent] = useMachine(printStateMachine)
+  const [printingState, setPrintingState] = useState('')
```
eventualbuddha: Use an enum for these values? Makes it less likely we accidentally typo them:

```
enum PrintingState {
  PrintBallot = 'PRINT_BALLOT'
  …
}
```

Names and values are up to you.

benadida: I need to do some fine tuning of this anyways for error conditions, so I will do that in a follow-on PR, but want to get a build going right now so merging prior to this fix.

PR #1160, opened 2020-02-01
Start screen instructions

beausmith:
Would be ideal to find the specific laws which apply to this content such that we can determine exactly what instructions are necessary here… if any.

image


eventualbuddha:
Not that it’ll change anything, but I wish they’d read stuff likes this: https://www.nngroup.com/articles/how-users-read-on-the-web/

> How Users Read on the Web
>
> Summary: They don't. People rarely read Web pages word by word; instead, they scan the page, picking out individual words and sentences.

beausmith:
@eventualbuddha - That article changed my life. I refer people to it often.

PR #1163, opened 2020-02-01
Candidate party not required on candidate contest.

beausmith:
`candidate.partyId` is no longer required on `CandidateContest`. This was already the case on `ReviewPage` and `PrintedBallot`.


eventualbuddha:
_(imitating broken record)_ test?

PR #1165, opened 2020-02-02
parallelize card clearing and tracker generation

benadida:
Deleting the card takes ~3-4s. Generating the tracker takes 4s. Let's put them in parallel and save ourselves 4s of printing dialogs.

Also, handle error conditions of no tracker generated and card not properly cleared.

### DIFF
--- a/src/pages/PrintOnlyScreen.tsx
+++ b/src/pages/PrintOnlyScreen.tsx
@@ -18,6 +18,17 @@ import { getBallotStyle, getPrecinctById } from '../utils/election'
 
 import encryptBallotWithElectionGuard from '../endToEnd'
 
+enum PrintingState {
+  StartPrinting = 1,
+  PrintBallot,
+  PrintTracker,
+  DonePrinting,
+  CardEmpty,
+  CardAbsent,
+  ErrorTracker,
+  ErrorCard,
+}
+
 const Graphic = styled.img`
   margin: 0 auto -1rem;
   height: 40vw;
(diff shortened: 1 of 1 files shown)


src/pages/PrintOnlyScreen.tsx
```diff
@@ -86,54 +97,89 @@ const PrintOnlyScreen = ({
...
           if (useElectionGuard) {
-            // no delay in next state because that state takes a while
-            setPrintingState('GENERATE_TRACKER')
-          } else {
-            printBallotTimer.current = window.setTimeout(() => {
-              setPrintingState('PRINT_DONE')
-            }, printingMessageTimeoutSeconds * 1000)
+            // don't await, just kick off both jobs
+            generateTrackingCode()
+            markCardUsed()
```
eventualbuddha: Seems like these should be `await`ed and the state change should happen after this. Something like this:

```ts
await Promise.all([generateTrackingCode(), markCardUsed()])
setPrintingState(…)
```

benadida: I think that's an alternative yes, but if you see how I handled the state and the error conditions for each one, I think my way is just as good and possibly simpler for error conditions?

eventualbuddha: I'm just concerned that nothing is handling the possible exceptions from those functions, but then again mine doesn't really either. In general I dislike seeing a promise-returning function that isn't `await`ed, but I think yours is fine for now.

benadida: that's fair, I can see how I'm abusing the async / await syntax a bit here. Let's discuss a better way! I'll merge into the branch and file a ticket.

eventualbuddha: My first impression: wrap calls in `try/catch` and do a `setError` or something in `catch`.

benadida: see #1166

PR #1172, opened 2020-02-02
reset state on printer disconnect

benadida:
This ensure that, when the printer is reconnected, we're not double-printing from stored state or anything sily.


benadida:
@beausmith I've removed the large comment and re-introduced a very small comment to explain the stop/start polling, which I think is important, and I think we can keep that comment in sync with those 2 lines of code :)

PR #1174, opened 2020-02-03
print ballot and tracker to specific trays

benadida:
Tray 3 is bigger, will have thicker paper, so used for ballots.

### DIFF
--- a/src/pages/PrintOnlyScreen.tsx
+++ b/src/pages/PrintOnlyScreen.tsx
@@ -82,7 +82,7 @@ const PrintOnlyScreen = ({
   }
 
   const printBallot = async () => {
-    await printer.print()
+    await printer.print('', 'Tray3')
     updateTally()
   }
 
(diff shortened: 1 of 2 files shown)


eventualbuddha:
> Nit: drop unused param, or pass obj with named props.

Object with named props is my plan for medium term (i.e. Monday or Tuesday, ideally). Otherwise, `undefined` would be better than `''` for `deviceName`.

benadida:
> > Nit: drop unused param, or pass obj with named props.
> 
> Object with named props is my plan for medium term (i.e. Monday or Tuesday, ideally). Otherwise, `undefined` would be better than `''` for `deviceName`.

I thought because this is message passing, undefined would be less than ideal? We'd have to change the kiosk browser code at least, I think, as it currently stands.

eventualbuddha:
You’re right @benadida. Carry on.

PR #1177, opened 2020-02-03
Print ballot regardless

benadida:
We could more aggressively detect a failed call, but we do want to also handle timeouts, so doing it all in one somewhat simple approach.


benadida:
> LGTM with a test or two.

Since we keep rev'ing the flow for `electionguard`, I want to stabilize it and then add the tests before we merge to `master`. I know, there's some risk here.

PR #1187, opened 2020-02-03
Display "printing" modal while report is printing.

### LINKED ISSUE #918: VxPrint: polls open / close report printing dialog stays up too long
Specifically, after the print dialog has appeared and disappeared, the dialog and buttons are still up. It misleads the user into thinking they have to press "yes" again. Not sure what happens if you do. Double print?

beausmith:
Closes #918


beausmith:
Simple fix to get this passing. 

Seems like the pre-commit hook should catch this. 😕

PR #1195, opened 2020-02-04
feat: remove app mode config from admin screen

eventualbuddha:
Instead, we fetch app  mode from `/machine-config` which pulls machine ID from `VX_MACHINE_ID` and app mode from `VX_APP_MODE` environment variables.

This will, for now, make it harder to switch between app modes in development. You'll have to restart the server with the right environment variable and run it like so:

```
$ VX_APP_MODE=VxPrint yarn start
```


src/AppRoot.tsx
```diff
@@ -181,7 +175,7 @@ class AppRoot extends React.Component<Props, State> {
     isFetchingElection: false,
     isLiveMode: false,
     isPollsOpen: false,
-    machineId: '---',
+    machineConfig: { appMode: VxMarkOnly, machineId: '0000' },
```
beausmith: nit: Drop one zero here so all the `productId`'s are three digits?

src/config/types.ts
```diff
@@ -15,18 +15,30 @@ export interface Provider<T> {
...
+  isVxMark: true,
+  isVxPrint: false,
+} as const
+export const VxMarkPlusVxPrint = {
   name: 'VxMark + VxPrint',
   isVxPrint: true,
   isVxMark: true,
+} as const
+export type AppMode =
+  | typeof VxMarkOnly
```
beausmith: Q: should this line start with a pipe?

src/utils/machineConfig.ts
```diff
@@ -0,0 +1,27 @@
...
+  Provider,
+  MachineConfig,
+  getAppMode,
+  AppModeNames,
+} from '../config/types'
+import fetchJSON from './fetchJSON'
+
+interface MachineConfigResponse {
+  machineId: string
+  appMode: string
```
beausmith: Should `appMode` be more tightly typed?

src/AppRoot.tsx
eventualbuddha: @benadida says this is typically 4  ‾\_(ツ)_/‾

src/config/types.ts
eventualbuddha: That's what prettier did. It's just a way to visually make it easier to grok.

src/AppRoot.tsx
beausmith: LOLz… what'evs. 😄

src/config/types.ts
beausmith: Cool. Good to know this syntax is acceptable. I like.

PR #1196, opened 2020-02-04
actually print blank ballots rather than assume the card is empty

benadida:


src/AppPrintOnly.test.tsx
```diff
@@ -38,6 +38,8 @@ beforeEach(() => {
 
 jest.useFakeTimers()
 
+jest.setTimeout(10000)
```
beausmith: I'm a little worried about increasing timeout without being certain that we know why.

src/AppPrintOnly.test.tsx
```diff
@@ -300,6 +302,39 @@ it('VxPrintOnly flow', async () => {
   // Check Printed Ballots Count
   getAllByTextWithMarkup('Ballots Printed: 3')
 
+  // Blank Ballot should print
```
beausmith: Lets be more specific: 
```suggestion
  // Card with no votes prints ballot with no votes
```

src/AppPrintOnly.test.tsx
```diff
@@ -300,6 +302,39 @@ it('VxPrintOnly flow', async () => {
...
+      isTestBallot: true,
+      ballotType: BallotType.Standard,
+    })
+  )
+
+  // Show Printing Ballot screen
+  await advanceTimersAndPromises()
+  getByText('Printing your official ballot')
+
+  expect(getAllByTextWithMarkup('[no selection]')).toHaveLength(20)
```
beausmith:
```suggestion
  // All contests should have no selection
  expect(getAllByTextWithMarkup('[no selection]')).toHaveLength(20)
```

beausmith: Wait… didn't this get changed to "no selection" without the brackets? Maybe this code is on the `electionguard` branch?

src/AppRoot.tsx
```diff
@@ -1043,7 +1045,7 @@ class AppRoot extends React.Component<Props, State> {
                   updateVote: this.updateVote,
                   forceSaveVote: this.forceSaveVote,
                   userSettings,
-                  votes,
+                  votes: votes ?? {},
```
beausmith: Since `{}` is used three times, lets add some context by adding a variable to the top of the file and reference it three times:
```ts
const EmptyVotes = {}
```

src/pages/PrintOnlyScreen.tsx
```diff
@@ -195,7 +194,8 @@ const PrintOnlyScreen = ({
           </MainChild>
         </Main>
       </Screen>
-      {isReadyToPrint && (
+      // && votes should not be required but tsc whines
```
beausmith: There is prob a better way to do this.

src/pages/PrintOnlyScreen.tsx
```diff
@@ -63,7 +62,7 @@ const PrintOnlyScreen = ({
   const printerTimer = useRef(0)
   const [okToPrint, setOkToPrint] = useState(true)
   const [isPrinted, updateIsPrinted] = useState(false)
-  const isCardVotesEmpty = isEmptyObject(votes)
+  const isCardVotesEmpty = !votes
```
beausmith: Won't this always be false? 🤔 

image

eventualbuddha:
I've got a fever, and the only prescription is more coverage.

image

src/AppRoot.tsx
```diff
@@ -746,7 +746,9 @@ class AppRoot extends React.Component<Props, State> {
         votes: { ...prevState.votes, [contestId]: vote },
       }),
       () => {
-        this.setVotes(this.state.votes)
+        if (this.state.votes) {
+          this.setVotes(this.state.votes)
+        }
```
beausmith: No "istanbul ignore line" ignore comment necessary if you use…
```
this.state.votes && this.setVotes(this.state.votes)
```

src/AppPrintOnly.test.tsx
benadida: it's just more printing :) That is all.

src/AppPrintOnly.test.tsx
benadida: I strongly prefer "blank ballot" because that is the technical term for a ballot without any options selected.

src/AppPrintOnly.test.tsx
benadida: This is on `master`, I don't think you've updated the ballot on `master`, right?

src/AppPrintOnly.test.tsx
beausmith:
```
  // Card with no votes prints a blank ballot with no votes
```
?

src/pages/PrintOnlyScreen.tsx
benadida: This is catching the case where `votes` is `undefined`. The big change in this PR is differentiating between `undefined` which means a card that hasn't been used, and `{}` which is a blank ballot.

beausmith:
```
const isCardVotesEmpty = votes === undefined
```
?

src/AppPrintOnly.test.tsx
benadida: That sounds incorrect to me --> a card with no votes sounds like a card that wasn't used. A card that contains a blank ballot should print, I maintain that the original comment is correct :)

eventualbuddha:
```
src/AppSetupErrors.test.tsx:154:23 - error TS2304: Cannot find name 'fakeMachineId'.

154     const machineId = fakeMachineId()
                          ~~~~~~~~~~~~~

src/AppSetupErrors.test.tsx:158:7 - error TS2345: Argument of type '{ appMode: { readonly name: "VxPrint"; readonly isVxMark: false; readonly isVxPrint: true; }; }' is not assignable to parameter of type 'Partial<State>'.
  Object literal may only specify known properties, and 'appMode' does not exist in type 'Partial<State>'.

158       appMode: VxPrintOnly,
          ~~~~~~~~~~~~~~~~~~~~

src/AppSetupErrors.test.tsx:165:9 - error TS2322: Type '{ card: MemoryCard; hardware: MemoryHardware; storage: MemoryStorage<AppStorage>; machineId: any; }' is not assignable to type 'IntrinsicAttributes & Props'.
  Property 'machineId' does not exist on type 'IntrinsicAttributes & Props'.

165         machineId={machineId}
            ~~~~~~~~~
```

`machineId` is now `machineConfig`. See other tests.

PR #1202, opened 2020-02-06
fix: keep machine config on unconfigure

### LINKED ISSUE #1198: VxPrint - unloading / reloading election switches machine back to VxMark mode

eventualbuddha:
Closes #1198

### DIFF
--- a/src/AppPrintOnly.test.tsx
+++ b/src/AppPrintOnly.test.tsx
@@ -32,6 +32,7 @@ import { MemoryHardware } from './utils/Hardware'
 import { fakeMachineConfigProvider } from '../test/helpers/fakeMachineConfig'
 import { REPORT_PRINTING_TIMEOUT_SECONDS } from './config/globals'
 import { VxPrintOnly } from './config/types'
+import { mockOf } from '../test/testUtils'
 
 beforeEach(() => {
   window.location.href = '/'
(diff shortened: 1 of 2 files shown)


src/AppPrintOnly.test.tsx
```diff
@@ -396,3 +397,106 @@ it('VxPrintOnly flow', async () => {
   // Default Unconfigured
   getByText('Device Not Configured')
 })
+
+test('VxPrint retains configuration after unload', async () => {
```
benadida: let's not call it "configuration", it should only retain machine type and machine ID.

eventualbuddha: Hardware configuration?

eventualbuddha: Hardware state?

eventualbuddha: How about "VxPrint retains app mode when unconfigured"?

src/AppPrintOnly.test.tsx
```diff
@@ -396,3 +397,106 @@ it('VxPrintOnly flow', async () => {
...
+
+  // Remove election configuration.
+  await unconfigure()
+
+  // Re-configure & open polls again.
+  await configure()
+  await openPolls()
+
+  // Make sure we're again ready to print ballots.
+  getByText('Insert Card to print your official ballot.')
```
beausmith: Suggestion: While we're abstracting to functions, this (and the line above) could be appended to the `openPolls` function.

eventualbuddha: I know, but I wanted it to be very clear what this test is checking.

PR #1209, opened 2020-02-06
fix: maybe address freezing issue

eventualbuddha:
This causes VxMark to quit after being idle on the "Insert Card" screen for 5 minutes. This MUST be paired with some sort of respawn functionality. It will only quit if the machine ID does _not_ end in "0", so with machine IDs using sequential integers this means ~90% of machines will have this restart behavior and  ~10% will not.

### DIFF
--- a/src/AppQuitOnIdle.test.tsx
+++ b/src/AppQuitOnIdle.test.tsx
@@ -1,0 +1,62 @@
+import * as React from 'react'
+import { render } from '@testing-library/react'
+
+import App from './App'
+
+import { advanceTimersAndPromises } from '../test/helpers/smartcards'
+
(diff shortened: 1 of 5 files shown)


benadida:
@eventualbuddha I'm wondering if it would be better to quit every X votes, rather than on idle time. Two reasons: idle time alone has never led to a freeze that we know of, only combination of some idle + active voting over at least a few hours, and a restart right after a voter removes their voter cards seems less likely to be perceived as a weird issue than if it happens right when a voter shows up at the 5 min idle mark.

I'm going to approve this now because it's good code and my concern is not that high, but I'd love your thoughts.

PR #1370, opened 2020-04-17
WIP: HMPB

beausmith:
This branch is functional for demo, but is very much a WIP as it still contains testing, other possibly unnecessary code, and has other limitations which should not exist in production code.


src/components/QRCode.tsx
```diff
@@ -10,13 +10,18 @@ const ResponsiveSvgWrapper = styled.div`
...
   value: string | Uint8Array
+  size?: number
+  bgColor?: string
+  fgColor?: string
+  level?: 'L' | 'M' | 'Q' | 'H'
+  renderAs?: 'svg' | 'canvas'
 }
 
-const QRCode = ({ value }: Props) => (
+const QRCode = ({ level = 'H', renderAs = 'svg', value }: QRCodeProps) => (
```
benadida, line 22: given that we know we're printing this and we know exactly how big it should be on paper, I wonder if we're better off rendering as a PNG given how SVG occasionally causes us trouble when printing.

src/App.tsx
```diff
@@ -43,6 +43,7 @@ const App = ({
   hardware = getHardware(),
   machineConfig = machineConfigProvider,
 }: Props) => {
+  // screenReader.disable()
```
beausmith, line 46: @eventualbuddha - Is there a better way to disable `screenReader` than adding uncommented version of this line?

src/components/QRCode.tsx
benadida: this was the printing issues with the Brother HL2300 and the SVG of the state seals. Not a blocker, just trying to play it defensive here since SVGs have been unfriendly to us in the past.

src/App.tsx
eventualbuddha: We should probably make it remember in local storage, or something.

src/components/QRCode.tsx
beausmith: That was 100% due to a horrible export to SVG by some wack tool. 

SVG will not be an issue for these QR codes.

beausmith:
Def not ready to merge. Comments are for testing and will be removed as @benadida guessed. 

Submitted as a draft PR (using little caret menu on web button to create a PR) assuming this is what this feature is for?

beausmith:
Next steps: 
1. Separate PagedJS renderer from PrintedBallot such that render happens once. 
2. Use post render hook to update the QR codes to contain page number and total pages for each voter ballot.

eventualbuddha:
Close?

PR #1445, opened 2020-07-08
fix: update ballot-encoder for new yes/no format

eventualbuddha:
I had to do this in order to generate updated fixtures for module-scan, but there's no urgency to review.


src/components/YesNoContest.tsx
```diff
@@ -252,10 +257,7 @@ export default class YesNoContest extends React.Component<Props> {
   }
 
   public closeOvervoteAlert = () => {
-    // Delay to avoid passing tap to next screen
-    window.setTimeout(() => {
-      this.setState({ overvoteSelection: initialState.overvoteSelection })
-    }, 200)
+    this.setState({ overvoteSelection: initialState.overvoteSelection })
```
benadida, line 260: I wouldn't want to remove the timeout right now unless we've done extensive testing. I know it *shouldn't* break, but this is super tricky.

eventualbuddha: Fair enough, though we did remove the other ones:
- e302f200b40ce1fa1977eb45ee939a1fdb8151ad
- 841cebe559c4bb80c9345377a76453ed405add44

src/utils/votes.ts
```diff
@@ -0,0 +1,9 @@
+import { YesNoVote } from '@votingworks/ballot-encoder'
+import { YesNo } from '../config/types'
```
beausmith, line 2: To make this even more clear…
```suggestion
import { YesOrNo } from '../config/types'
```

beausmith:
What's status of this PR?

src/utils/votes.ts
eventualbuddha: Fixed.

beausmith (approved):

PR #1486, opened 2020-09-21
feat: image support in yes/no contest description.

beausmith:
**Known issue:** the logic to display the "See More" buttons runs before the image is resized and thus the "See More" buttons are displayed when they shouldn't be. We should fix this but not currently a blocker.

## Screenshot
image


src/components/Text.tsx
```diff
@@ -73,7 +74,12 @@ export const TextWithLineBreaks = ({ text }: { text: string }) => (
       <p key={x}>
         {x.split(/[\n\r]/g).map((y, i, arr) => (
           <React.Fragment key={y}>
-            {y}
+            <span
+              // eslint-disable-next-line react/no-danger
+              dangerouslySetInnerHTML={{
+                __html: DOMPurify.sanitize(y),
+              }}
+            />
```
benadida, line 82: this turns on markup for all `Text`. No concerns?

benadida (approved):
just a question to make sure we're not missing any problematic consequences of enabling markup on all `Text`, but otherwise good to go.

eventualbuddha (approved):

src/components/Text.tsx
beausmith: None for now.

benadida (approved):

PR #1487, opened 2020-09-21
Fix/update ballot encoder with new yesno format2

beausmith:
This is https://github.com/votingworks/bmd/pull/1445 rebased with an additional commit.


eventualbuddha:
Obviated by #1445.

PR #1494, opened 2020-10-01
Feat/support ms either neither

beausmith:
Tasks to do before this is ready to merge… 
- [x] move the sample `ms-either-neither` contest to `ballot-encoder`. Pending: https://github.com/votingworks/ballot-encoder/pull/44
- [x] I need to review how this is represented on the BMD ballot.
- [x] Review audio version. Add pauses as necessary.
- [x] Test coverage for printed ballot variations of Either-Neither.

Also note that there is some code to placate the BMD tally code temporarily until that feature lands after this one.

image

image


beausmith:
Tests should pass… but there are a few lines without test coverage in the PrintedBallot component. Shouldn't be much work to fix.

benadida (approved):

benadida (approved):

PR #1497, opened 2020-10-03
fill in test deck ballots for either-neither and test them

benadida:


src/pages/TestBallotDeckScreen.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+      appPrecinctId="23"
+      election={electionSample as Election}
+      hideTestDeck={jest.fn()}
+      isLiveMode={false}
+    />
+  )
+
+  fireEvent.click(getByText('All Precincts'))
+
+  expect(container.firstChild).toMatchSnapshot()
```
beausmith, line 32: We should not use snapshots when there is data which we can find and verify on the rendered screen. It is too easy for someone to just update the snapshots and assume that everything is fine.

src/pages/TestBallotDeckScreen.tsx
```diff
@@ -58,10 +58,7 @@ const generateTestDeckBallots = ({
             ? 2
             : c.type === 'candidate'
             ? (c as CandidateContest).candidates.length
-            : /* istanbul ignore next - TODO: Tally: Remove this line once tally code is completed */
-            c.type === 'ms-either-neither'
-            ? 9 // TODO: Tally: 9 because: no vote + A + B
-            : 0
+            : 2 // ms-either-neither
```
beausmith, line 61: Else case as 'ms-either-neither` worries me. I feel like this is gonna come back to bite us if it's not an explicit contest type check.

beausmith (approved):
Approving… but ideally the two suggested fixes are made before merging. However if this cannot be done due to lack of time, please add a TODO in the code with links to tickets to fix these items.

src/pages/TestBallotDeckScreen.test.tsx
benadida: I generally agree with you, but in this case I was trying to capture the actual test deck, which I think we may want to have remain stable? (I think I may have captured the wrong snapshot, so I'll review that.)

Does your opinion change if the goal here is to capture the whole test deck and notice if it ever changes?

src/pages/TestBallotDeckScreen.tsx
benadida: Yeah, I did this because the other approach then caused a code coverage problem, but I can `istanbul` ignore that last line. will change.

src/pages/TestBallotDeckScreen.test.tsx
benadida: never mind, I found some things I can test in the content that address both issues, so all good. Thanks for the nudge!

PR #1498, opened 2020-10-06
feat: Admin setting for date, time, and zone.

beausmith:
This uses select elements for all settings. 

bmd-datetime-settings3


benadida (approved):
looks great.

src/config/globals.ts
```diff
@@ -18,3 +18,12 @@ export enum YES_NO_VOTES {
 export const WRITE_IN_CANDIDATE_MAX_LENGTH = 40
 export const LOW_BATTERY_THRESHOLD = 0.25
 export const QUIT_KIOSK_IDLE_SECONDS = 5 * 60 // 5 minutes
+export const AMERICA_TIMEZONES = [
+  'America/Honolulu',
+  'America/Anchorage',
+  'America/Los_Angeles',
+  'America/Phoenix',
+  'America/Salt_Lake_City',
```
eventualbuddha, line 26: I don't think this is a real IANA timezone. It's not in the [database](https://www.iana.org/time-zones) and [time.is says SLC's IANA identifier is America/Denver](https://time.is/Salt_Lake_City). I think we should replace this with `America/Denver`.

```suggestion
  'America/Denver',
```

Also, I think it's worth considering using different labels for these, like "Hawaiian Time", "Pacific Time", "Eastern Time", etc.

src/pages/ClerkScreen.tsx
```diff
@@ -39,15 +54,50 @@ const ClerkScreen = ({
...
+    const [year, month, day] = event.currentTarget.value.split('-')
+    setSystemDate(
+      new Date(
+        parseInt(year),
+        parseInt(month) - 1,
+        parseInt(day),
+        systemDate.getHours(),
+        systemDate.getMinutes()
+      )
+    )
```
eventualbuddha, line 78:
```suggestion
    const datePart = new Date(event.currentTarget.value)
    setSystemDate(
      datePart.getFullYear(),
      datePart.getMonth(),
      datePart.getDate(),
      systemDate.getHours(),
      systemDate.getMinutes()
    )
```

src/pages/ClerkScreen.tsx
```diff
@@ -39,15 +54,50 @@ const ClerkScreen = ({
...
+  const updateSystemTime: InputChangeEventFunction = (event) => {
+    const [hours, minutes] = event.currentTarget.value.split(':')
+    setSystemDate(
+      new Date(
+        systemDate.getFullYear(),
+        systemDate.getMonth(),
+        systemDate.getDay(),
+        parseInt(hours),
+        parseInt(minutes)
+      )
```
eventualbuddha, line 89:
```suggestion
      new Date(
        systemDate.getFullYear(),
        systemDate.getMonth(),
        systemDate.getDate(),
        parseInt(hours, 10),
        parseInt(minutes, 10)
      )
```

`Date#getDay()` returns the day of the week, not the day of the month.

src/pages/ClerkScreen.tsx
```diff
@@ -39,15 +54,50 @@ const ClerkScreen = ({
...
+    )
+  }
+  const updateTimeZone: SelectChangeEventFunction = (event) => {
+    setSystemTimeZone(event.currentTarget.value)
+  }
+  const setDateAndZone = () => {
```
eventualbuddha, line 95: Call this something like `commit` or `save` instead of `set` to clearly separate it from local state?

src/pages/ClerkScreen.tsx
```diff
@@ -188,6 +251,56 @@ const ClerkScreen = ({
...
+                  value={systemTimeZone}
+                  onBlur={updateTimeZone}
+                  onChange={updateTimeZone}
+                >
+                  <option value="" disabled>
+                    Select timezone…
+                  </option>
+                  {AMERICA_TIMEZONES.map((timezone) => (
+                    <option key={timezone} value={timezone}>
+                      {timezone.replace(/_/gi, ' ')}
```
eventualbuddha, line 287: If we're displaying without underscores here, should we do that on the main screen too?

src/pages/ClerkScreen.tsx
```diff
@@ -188,6 +251,56 @@ const ClerkScreen = ({
...
+              </InputGroup>
+            </div>
+          </Prose>
+        }
+        actions={
+          <React.Fragment>
+            <Button primary onPress={setDateAndZone}>
+              Save
+            </Button>
+            <Button onPress={() => setIsDateModalActive(false)}>Cancel</Button>
```
eventualbuddha, line 300: Should this reset the local state? i.e. `setSystemDate(new Date())`

src/config/types.ts
```diff
@@ -63,6 +63,16 @@ export function getAppMode(name: AppModeNames): AppMode {
...
+type ReactChangeEventFunction<T> = (event: React.ChangeEvent<T>) => void
+export type InputChangeEventFunction = ReactChangeEventFunction<
+  HTMLInputElement
+>
+export type TextareaChangeEventFunction = ReactChangeEventFunction<
+  HTMLTextAreaElement
+>
+export type SelectChangeEventFunction = ReactChangeEventFunction<
+  HTMLSelectElement
+>
```
eventualbuddha, line 75:
```suggestion
export type InputChangeEventFunction = React.ChangeEventHandler<
  HTMLInputElement
>
export type TextareaChangeEventFunction = React.ChangeEventHandler<
  HTMLTextAreaElement
>
export type SelectChangeEventFunction = React.ChangeEventHandler<
  HTMLSelectElement
>
```

eventualbuddha (approved):
LGTM after some minor fixes. The only thing that has to change is the `getDay` bug.

src/config/globals.ts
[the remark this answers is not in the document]
beausmith: Yes. Arizona is hipster. 
image

beausmith: Yes. Let's use more friendly names.

src/pages/ClerkScreen.tsx
beausmith: Good call.

src/pages/ClerkScreen.tsx
beausmith: yup!

src/pages/ClerkScreen.tsx
beausmith: yep, but gonna move to timezone names.

src/config/globals.ts
```diff
@@ -18,3 +18,26 @@ export enum YES_NO_VOTES {
...
+  'Apr',
+  'May',
+  'Jun',
+  'Jul',
+  'Aug',
+  'Sep',
+  'Oct',
+  'Nov',
+  'Dec',
+]
```
eventualbuddha, line 43: Take it or leave it, I felt compelled to see what it'd look like:

```suggestion
function* getShortMonthNames(): Generator<string> {
  const monthShortNameFormatter = new Intl.DateTimeFormat(undefined, {
    month: 'short',
  })
  const year = new Date().getFullYear()

  for (
    let month = 0;
    new Date(year, month, 1).getFullYear() === year;
    month++
  ) {
    yield monthShortNameFormatter.format(new Date(year, month, 1))
  }
}

export const MONTHS_SHORT = [...getShortMonthNames()]
```

src/pages/ClerkScreen.tsx
```diff
@@ -39,15 +56,77 @@ const ClerkScreen = ({
...
   const showTestDeck = () => setIsTestDeck(true)
   const hideTestDeck = () => setIsTestDeck(false)
+
+  const [isSavingDate, setIsSavingDate] = useState(false)
+  const [isSystemDateModalActive, setIsSystemDateModalActive] = useState(false)
+  const [systemDate, setSystemDate] = useState(new Date())
+  const [systemMeridian, setSystemMeridan] = useState<Meridian>(
+    systemDate.getHours() < 12 ? 'AM' : 'PM'
+  )
+  const [systemTimezoneLabel, setSystemTimezoneLabel] = useState(
```
eventualbuddha, line 73: Why store the label? Just store the IANA value and look up the label.

src/pages/ClerkScreen.tsx
```diff
@@ -188,6 +280,172 @@ const ClerkScreen = ({
...
+                    disabled={isSavingDate}
+                    onBlur={updateSystemTime}
+                    onChange={updateSystemTime}
+                  >
+                    <option value="" disabled>
+                      Year
+                    </option>
+                    {[...Array(11).keys()].map((year) => (
+                      <option key={year} value={year + 2020}>
+                        {year + 2020}
```
eventualbuddha, line 307:
```suggestion
                      <option key={year} value={year + new Date().getFullYear()}>
                        {year + new Date().getFullYear()}
```

src/config/globals.ts
```diff
@@ -18,3 +18,26 @@ export enum YES_NO_VOTES {
 export const WRITE_IN_CANDIDATE_MAX_LENGTH = 40
 export const LOW_BATTERY_THRESHOLD = 0.25
 export const QUIT_KIOSK_IDLE_SECONDS = 5 * 60 // 5 minutes
+export const AMERICA_TIMEZONES = [
```
eventualbuddha, line 21: Should this stuff go in `utils/date.ts`?

eventualbuddha (approved):

src/pages/ClerkScreen.tsx
beausmith: If the year was set to 2030, then with this it would be impossible to change the year back to 2020, right?

src/config/globals.ts
beausmith: Sounds reasonable. WIll do.

src/pages/ClerkScreen.tsx
eventualbuddha: Ah, good point. Carry on.

src/pages/ClerkScreen.tsx
beausmith: Good refactor. Done.

src/config/globals.ts
beausmith: Lets do it.

benadida (approved):
Looks good to me, with one issue: why are we showing "Standard Time" in the dropdown? That implies that they are picking daylight savings or not, which they're not, and it could get confusing. Any way to not have it say "standard" or "daylight"?

If it's too hard, I can absolutely live with this for now, and in fact I would urge you to merge anyways and tweak later, as the underlying functionality is as desired, this is just a display tweak potentially.

eventualbuddha (approved):

eventualbuddha:
@benadida the label is dynamic and generated by the system, so it's always correct:

ezgif com-optimize-2

PR #1502, opened 2020-10-12
fix muting logic

benadida:
As best as I can tell, we've had it backwards for a while. We were muting when an accessible controller was plugged in.


eventualbuddha (approved):
Yep, c9e5b0204e281b0ac8aa3dfcc6b0fbf22cbe720e broke it.

PR #1503, opened 2020-10-12
don't send extra args to kiosk.print for test deck

benadida:
doing so appears to crash kiosk, likely because we're sending an event and it's expecting an options object.

we should fix kiosk at some point, but simplest fix right now is to not send anything to `kiosk.print()` unless we mean to.


eventualbuddha (approved):

beausmith (approved):

beausmith:
Man! This has bitten me before 🤦‍♂️

benadida:
> Man! This has bitten me before

And to be clear this is because kiosk-browser changed from under `bmd`, so ... just a sign of us needing to figure out better full integration testing.

PR #1511, opened 2020-10-19
fix(coverage): update istanbul-reports to fix HTML coverage reports

eventualbuddha:
Per https://github.com/facebook/jest/issues/9388#issuecomment-573371732.


beausmith (approved):
Boom!

